PR #1, opened 2019-06-20
Add documentation

rylanc:


README.md
```diff
@@ -18,6 +18,131 @@ Or install it yourself as:
...
+class User < BaseObject
+  def self.resolve_reference(reference, context)
+    USERS.find { |user| user[:id] == reference[:id] }
+  end
+end
+```
+
+## Known Limitations and Issues
+ - Currently only works with class-based schemas
+ - Does add directives to the output of `Schema.to_definition`. Since `graphql-ruby` doesn't natively support schema directives, the directives will only be visible to the [Apollo Gateway](https://www.apollographql.com/docs/apollo-server/api/apollo-gateway/) through the `Query._service` field (see the [Apollo Federation specification](https://www.apollographql.com/docs/apollo-server/federation/federation-spec/))
```
noaelad: @rylanc missing a `not` in this sentence ("Does not add directives")

PR #5, opened 2019-06-27
Add CI

rylanc:


.travis.yml
```diff
@@ -0,0 +1,23 @@
+matrix:
+  include:
+  - name: 'rspec Tests'
+    language: ruby
+    rvm:
+      - 2.3
+    cache: bundler
+    before_install:
+      - gem update --system
+      - gem install bundler
```
noaelad: I'm curious - how does travis know to run `bundle exec rake`? Is that just their default for running ruby tests? If we override that with an `rspec` command could we avoid the rake dependency?

rylanc: Yeah, it just runs rake by default. I think it's kinda nice to have a `Rakefile`, since I think that's how most gems handle running tests (it's test framework agnostic).

noaelad: That's fair, let's keep it

noaelad:
This looks great, thanks for setting it up!

zp-alfred:
Accepted by noaelad!

PR #6, opened 2019-06-27
chore: Add rubocop

rylanc:
- Adds rubocop
 - Autocorrects all fixable offenses


.travis.yml
```diff
@@ -8,6 +8,9 @@ matrix:
     before_install:
       - gem update --system
       - gem install bundler
+    script:
+      - bundle exec rake
+      - bundle exec rake rubocop
```
noaelad: Have you tested that this breaks when you introduce a new rubocop violation?

rylanc: Yup. Here's an example: https://travis-ci.org/Gusto/apollo-federation-ruby/jobs/551943556

zp-alfred:
Accepted by noaelad!

PR #10, opened 2019-07-11
feat: Add description to the _Service type

rylanc:


lib/apollo-federation/service.rb
```diff
@@ -5,6 +5,9 @@
 module ApolloFederation
   class Service < GraphQL::Schema::Object
     graphql_name '_Service'
+    description 'The sdl representing the federated service capabilities. Includes federation ' \
```
bagelbits: Should `sdl` be capitalized?

zp-alfred:
Accepted by bagelbits!

lib/apollo-federation/service.rb
rylanc: I copied the description directly from the JS `@apollo/federation` lib: https://github.com/apollographql/apollo-server/blob/master/packages/apollo-federation/src/types.ts#L28

PR #12, opened 2019-07-28
Handle rails params in Any

tomasz-buchta:
Using with rails causes NoMethodError as the value is an instance of rails params and they don't implement each_key method (when using default scaffolded GraphQLController)

### DIFF
--- a/lib/apollo-federation/any.rb
+++ b/lib/apollo-federation/any.rb
@@ -9,8 +9,8 @@ module ApolloFederation
     def self.coerce_input(value, _ctx)
       # TODO: Should we convert it to a Mash-like object?
       result = {}
-      value.each_key do |key|
-        result[key.to_sym] = value[key]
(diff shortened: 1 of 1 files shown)


tomasz-buchta:
Let me know if a reproduction repo/more specs are needed

rylanc:
Thanks for the PR. What version of Rails and graphql-ruby are you using?

tomasz-buchta:
1st service: rails 5.2.3  graphql 1.9.4
2nd service: rails 5.2.3 graphql 1.9.8

I can try to update them, but both those services are running fine with patch from this pr. Without it the federation cannot resolve references.

lennyburdette:
I got the same error with Rails 5.2.3 and graphql 1.9.7.

zp-alfred:
Changes requested by rylanc.

rylanc:
Closing in favor of #13. Thanks for the work!

PR #13, opened 2019-07-31
fix: Handle rails params in Any 

lennyburdette:
Supersedes #12 to include tests

cc @tomasz-buchta @rylanc


zp-alfred:
Accepted by rylanc!

PR #15, opened 2019-08-01
chore: Add semantic-release

rylanc:
Enable easy gem releases.

I couldn't think of a good way to test this. I've tested it locally, without actually pushing the gem and things work OK up to that point.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -32,7 +32,5 @@ jobs:
       #   - make
       deploy:
         provider: script
-        skip_cleanup: true
-        script:
-          # Use nvm to install and use the Node LTS version (nvm is installed on all Travis images)
-          - echo 'RELEASE'
+        # skip_cleanup: true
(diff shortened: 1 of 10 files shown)


.travis.yml
```diff
@@ -1,26 +1,36 @@
...
+      before_install:
+        - rvm use 2.3 --install --binary --fuzzy
+        - ruby --version
+        - gem update --system
+        - gem install bundler
+        - bundle install --jobs=3 --retry=3 --deployment
+      script:
+        - yarn test
+        - yarn lint
+    # TODO: Only add this stage on master?
```
bagelbits: It probably makes sense to only add this stage on master. I think you can add this:
```
branches:
  only:
  - master
```

rylanc: Done!

zp-alfred:
Accepted by bagelbits!

PR #17, opened 2019-08-11
fix: Support lazy objects

cmschuetz:
This introduces a mechanism for resolving lazy objects before setting their type on the query context.  Previously, type resolution didn't work because the lazy object was being set instead of its resolved value.

I believe `after_resolve` is the correct hook to introduce this behavior as objects will be 'synced', but curious if there are other strategies for accomplishing this


DamirSvrtan:
Great work @cmschuetz! @rylanc anything blocking from merging this in?

cmschuetz:
fixed the builds

prikha:
Tested on a small artificial app with https://github.com/shopify/graphql-batch - works. 

@noaelad @rylanc could you find some time to review and merge it.

noaelad:
Apologies for the delay! I'll take a look at it today

lib/apollo-federation/entity_type_resolution_extension.rb
```diff
@@ -0,0 +1,18 @@
+# frozen_string_literal: true
+
+class EntityTypeResolutionExtension < GraphQL::Schema::FieldExtension
+  def after_resolve(value:, context:, **_rest)
+    synced_value =
+      value.map do |type, result|
+        [type, context.query.schema.sync_lazy(result)]
```
noaelad: I'm a little confused about the need to call `sync_lazy` in `after_resolve`, from reading [the docs](https://graphql-ruby.org/type_definitions/field_extensions.html#modifying-field-configuration) it seems like `after_resolve` should have been called after all the lazy objects have synced, however that does not seem to be the case. I pulled this branch down and played with it a bit and it does seem like the objects haven't been "lazy loaded" until this point, so I agree this is needed but just not sure why.

cmschuetz: I also initially thought that the lazy values would be resolved at this point which led me to reach for this solution.  My best guess is that the docs mean that there are are no more lazy evaluations in flight for the given field.  E.g. In the context of `graphql-batch`, all keys have been collected for the given loader and it's 'ready' to be resolved.

PR #18, opened 2019-08-15
feat: switch from Travis to CircleCI

grxy:
Co-authored-by: Grex <grex@gusto.com>
Co-authored-by: Noa Elad <noa.elad@gusto.com>


.circleci/config.yml
```diff
@@ -0,0 +1,38 @@
+version: 2.1
+
+orbs:
+  gusto: gusto/gusto@volatile
```
grxy: We probs want to explicitly pick a version here and then auto-PR updates with renovate. Orb is here: https://circleci.com/orbs/registry/orb/gusto/gusto

grxy:
@noaelad This should now be code-complete. We just need to add the correct env vars to Circle so that we can auto-publish correctly.

PR #26, opened 2019-09-17
Fix typo on exception message

jmonteiro:


rylanc:
Thanks for the contribution, but this has been fixed in https://github.com/Gusto/apollo-federation-ruby/pull/33

PR #27, opened 2019-09-18
feat: Support Interfaces

rtymchyk:
Following the spec defined in: https://www.apollographql.com/docs/apollo-server/federation/federation-spec/

**Notes:**
`directive @extends on OBJECT | INTERFACE`
`directive @key(fields: _FieldSet!) on OBJECT | INTERFACE`

-> Introduce a new `ApolloFederation::Interface` module that can be included in base interface modules, using the API outlined in https://graphql-ruby.org/type_definitions/interfaces#definition-methods

`union _Entity`
Unions cannot reference interface types, so the resolver excludes non-object types to generate this field.

### DIFF
--- a/README.md
+++ b/README.md
@@ -44,6 +44,18 @@ class BaseObject < GraphQL::Schema::Object
 end
 ```
 
+Include the `ApolloFederation::Interface` module in your base interface module:
+
+```ruby
+include GraphQL::Schema::Interface
+  field_class BaseField
+
+  definition_methods do
+    include ApolloFederation::Interface
+  end
+end
+```
+
 Finally, include the `ApolloFederation::Schema` module in your schema:
 
 ```ruby
--- a/README.md
+++ b/README.md
@@ -49,12 +49,9 @@ Include the `ApolloFederation::Interface` module in your base interface module:
 ```ruby
 module BaseInterface
   include GraphQL::Schema::Interface
+  include ApolloFederation::Interface
 
   field_class BaseField
-
-  definition_methods do
-    include ApolloFederation::Interface
-  end
 end
 ```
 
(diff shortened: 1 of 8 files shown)


rtymchyk:
Having a rough time adding new tests because adding modules as fixtures seems to be a pain

rtymchyk:
Any early thoughts @rylanc?

rtymchyk:
I've added tests, this should be ready to go. @rylanc @grxy

rylanc:
Sorry for the delay. I've been out on parental leave for the last few weeks. I'll take a look at this next week.

rtymchyk:
Thank you for the update 🙂

README.md
```diff
@@ -44,6 +44,20 @@ class BaseObject < GraphQL::Schema::Object
...
 
+Include the `ApolloFederation::Interface` module in your base interface module:
+
+```ruby
+module BaseInterface
+  include GraphQL::Schema::Interface
+
+  field_class BaseField
+
+  definition_methods do
```
rylanc: This interface seems a little clunky and inconsistent. We could make it so you simply `include ApolloFederation::Interface` in the module definition by adding the definition methods in an `included` hook:
```ruby
module ApolloFederation
  module Interface
    def self.included(klass)
      klass.definition_methods do
        include DefinitionMethods
      end
    end

    module DefinitionMethods
      include HasDirectives

      def extend_type
        add_directive(name: 'extends')
      end

      def key(fields:)
        add_directive(
          name: 'key',
          arguments: [
            name: 'fields',
            values: fields,
          ],
        )
      end
    end
  end
end
```

What do you think?

rtymchyk: Ok makes sense, I'll see if I can make the change.

rtymchyk: @rylanc done

rtymchyk: Aaand linted 😁

PR #31, opened 2019-10-04
chore: upgrade orb and use semantic-release-ruby command

grxy:
This PR upgrades the CI config to use the newly-added `semantic-release-ruby` orb command.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,7 +1,7 @@
 version: 2.1
 
 orbs:
-  gusto: gusto/gusto@0.0.9
+  gusto: gusto/gusto@0.0.10
 
 jobs:
   ruby-test:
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -19,14 +19,7 @@ jobs:
...
     steps:
-      - gusto/bundle-install
-      - gusto/yarn-install
-      - run:
-          command: |
-            mkdir -p ~/.gem
-            echo -e "---\r\n:rubygems_api_key: $GEM_HOST_API_KEY" > ~/.gem/credentials
-            chmod 0600 /home/circleci/.gem/credentials
-      - run: npx semantic-release --debug
+      - gusto/semantic-release-ruby
```
noaelad: I think we still want to keep the bundle-install and yarn-install steps, no?

.circleci/config.yml
```diff
@@ -19,14 +19,7 @@ jobs:
   release:
     executor: gusto/ruby-2-3
     steps:
-      - gusto/bundle-install
-      - gusto/yarn-install
-      - run:
-          command: |
-            mkdir -p ~/.gem
-            echo -e "---\r\n:rubygems_api_key: $GEM_HOST_API_KEY" > ~/.gem/credentials
-            chmod 0600 /home/circleci/.gem/credentials
```
noaelad: And this third command also looks like we should keep it around

.circleci/config.yml
grxy: These were all rolled up into a shared command here: https://github.com/Gusto/circleci-orbs/blob/master/orbs/gusto/commands/semantic-release-ruby.yml

.circleci/config.yml
grxy: These were all rolled up into a shared command here: https://github.com/Gusto/circleci-orbs/blob/master/orbs/gusto/commands/semantic-release-ruby.yml

PR #32, opened 2019-10-08
docs: Clarify intent not to support legacy define API

daemonsy:
## What's up?

On the graphql-ruby [roadmap](https://graphql-ruby.org/schema/class_based_api.html#roadmap), the `.define` API is marked for removal at `2.0` and sunsetted at `1.10` (which is the upcoming release). 

Hence just a small README change to clarify that we won't do any work to support the legacy API.

### DIFF
--- a/README.md
+++ b/README.md
@@ -172,7 +172,7 @@ To support [federated tracing](https://www.apollographql.com/docs/apollo-server/
     ```
 
 ## Known Issues and Limitations
- - Currently only works with class-based schemas
+ - Only works with class-based schemas, the legacy `.define` API will not be supported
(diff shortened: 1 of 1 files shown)


rtymchyk:
I think we should clarify that this gem is still usable, as long as the schema definition and the graph types that you attach it to are class based.

rylanc:
@rtymchyk Do you mean that if there are types that don't use federation are defined using `.define`, everything will still work? I haven't tried it myself, but it makes sense.

rtymchyk:
Yeah @rylanc that was the incremental approach I took. It's only necessary if you use the APIs you defined here + for query/mutation root types.

daemonsy:
Thank you!

PR #34, opened 2019-10-11
fix: drop actionpack from runtime requirements

prikha:
At this moment distributed tracing has an `active_support` runtime dependency. Which might not be welcome in some Ruby projects. 

It would be nice to leave it at least a development dependency or drop completely.

### DIFF
--- a/lib/apollo-federation/tracing/node_map.rb
+++ b/lib/apollo-federation/tracing/node_map.rb
@@ -1,6 +1,5 @@
 # frozen_string_literal: true
 
-require 'active_support/core_ext/array/wrap'
 require 'apollo-federation/tracing/proto'
 
 module ApolloFederation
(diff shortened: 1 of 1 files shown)


prikha:
@noaelad would you mind taking a look at currently open PR's. There are a few including this that remove friction for those playing around with the library.

prikha:
There is spec for `ActionController::Param` coercion if I am not mistaken.

prikha:
Noa @noaelad can you give me a hint how to trigger the builds though. I feel like the spec is legit. Are you sure we want to drop it?

noaelad:
@prikha I think you just need to login to Circle CI with your github account and maybe set it up for your fork of this repo. This PR is an example of a fork running the test suite:
https://github.com/Gusto/apollo-federation-ruby/pull/16
Let me know if you run into trouble getting that working

noaelad:
You are right about the spec, sorry I missed that one. Let's keep it as a dev dependency then.

prikha:
@noaelad what do I need to do to:
a) make CI run
b) have an approval and merge the PR
?

PR #35, opened 2019-10-11
fix: specify dependency versions

prikha:
BREAKING CHANGE: Requires graphql ~> 1.9.8 and google-protobuf ~> 3.7

There is no version specified for the dependencies. This is unpleasant for those trying to integrate the gem with existing codebase. Even though the library is its early development stage it would be nice to at least specify requirements. Versions could be adjusted later on, but it's way easier to spot a problem that to see misleading `protobuf` errors.

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -2,8 +2,8 @@ PATH
   remote: .
   specs:
     apollo-federation (0.5.1)
-      google-protobuf
-      graphql
+      google-protobuf (~> 3.7)
+      graphql (~> 1.9.8)
 
 GEM
   remote: https://rubygems.org/
--- a/apollo-federation.gemspec
+++ b/apollo-federation.gemspec
@@ -26,9 +26,9 @@ Gem::Specification.new do |spec|
 
   spec.files = `git ls-files bin lib *.md LICENSE`.split("\n")
 
-  spec.add_dependency 'graphql'
+  spec.add_dependency 'graphql', '~> 1.9.8'
 
-  spec.add_runtime_dependency 'google-protobuf'
+  spec.add_runtime_dependency 'google-protobuf', '~> 3.7'
 
   spec.add_development_dependency 'actionpack'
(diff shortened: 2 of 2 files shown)


rylanc:
@lennyburdette Any idea what the minimum version of `google-protobuf` is that would work with federated tracing?

lennyburdette:
Using `~> 3.0` should work just fine, but I wouldn't say that google-protobuf uses strict semantic versioning. Glancing at the changelog, they've [dropped support for older versions of Ruby](https://github.com/protocolbuffers/protobuf/releases/tag/v3.7.0) and [removed API](https://github.com/protocolbuffers/protobuf/releases/tag/v3.9.0) in minor releases.

prikha:
@lennyburdette @rylanc I get what you say, but the PR came from a need: 

Version **3.6.0** clearly doesn't work
```
Failure/Error:
  add_file('apollo.proto', syntax: :proto3) do
    add_message 'mdg.engine.proto.Trace' do
      optional :start_time, :message, 4, 'google.protobuf.Timestamp'
      optional :end_time, :message, 3, 'google.protobuf.Timestamp'
      optional :duration_ns, :uint64, 11
      optional :root, :message, 14, 'mdg.engine.proto.Trace.Node'
      optional :signature, :string, 19
      optional :details, :message, 6, 'mdg.engine.proto.Trace.Details'
      optional :client_name, :string, 7
      optional :client_version, :string, 8

NoMethodError:
  undefined method `add_file' for #<Google::Protobuf::Internal::Builder:0x00007fbb4ca90f10>
  Did you mean?  add_enum
```

Version **3.7.0** and above seems to be good. So changed the minimum required version to **3.7.0**

prikha:
@lennyburdette @rylanc @noaelad ☝️

prikha:
@lennyburdette @rylanc @noaelad ☝️

prikha:
@rylanc done

PR #45, opened 2020-01-03
fix: Send context to SDL generation / don't cache

cmschuetz:
Recently, `graphql-ruby` started supporting visibility rules during SDL generation (PR: https://github.com/rmosolgo/graphql-ruby/pull/2637).  In order to utilize dynamic values set in the query context, it needs to be forwarded to `FederatedDocumentFromSchemaDefinition`.  Additionally, the current caching mechanism poses a problem if you have several gateways/clients with different visibility rules, so it has been removed.

### DIFF
--- a/lib/apollo-federation/schema.rb
+++ b/lib/apollo-federation/schema.rb
@@ -47,7 +47,7 @@ module ApolloFederation
         super
       end
 
-      def federation_sdl(context)
+      def federation_sdl(context: nil)
         document_from_schema = FederatedDocumentFromSchemaDefinition.new(self, context: context)
         GraphQL::Language::Printer.new.print(document_from_schema.document)
       end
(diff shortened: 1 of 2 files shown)


coontail:
Looks amazing !

cmschuetz:
@grxy @kryptonat @rylanc @noaelad anyone have thoughts on this?

lib/apollo-federation/schema.rb
```diff
@@ -47,11 +47,9 @@ def to_graphql
         super
       end
 
-      def federation_sdl
-        @federation_sdl ||= begin
-          document_from_schema = FederatedDocumentFromSchemaDefinition.new(self)
-          GraphQL::Language::Printer.new.print(document_from_schema.document)
-        end
+      def federation_sdl(context)
```
grxy: Can we default this argument to `nil` to prevent a breaking change?

cmschuetz:
@grxy Updated, thanks for the review

noaelad:
Thanks for adding this!

PR #50, opened 2020-01-29
fix: service field context

matthewjf:
Methods definition here:
https://github.com/Gusto/apollo-federation-ruby/blob/master/lib/apollo-federation/schema.rb#L50

Currently we get an exception when running the service check command:
```
apollo schema:check --endpoint=http://localhost:3000/graphql ...
```
Results in:
```
unknown keyword: current_user
/Users/matt/.rbenv/versions/2.6.5/lib/ruby/gems/2.6.0/gems/apollo-federation-1.0.1/lib/apollo-federation/schema.rb:50:in `federation_sdl'
/Users/matt/.rbenv/versions/2.6.5/lib/ruby/gems/2.6.0/gems/apollo-federation-1.0.1/lib/apollo-federation/service_field.rb:13:in `_service'
/Users/matt/.rbenv/versions/2.6.5/lib/ruby/gems/2.6.0/gems/graphql-1.9.19/lib/graphql/schema/field.rb:652:in `public_send'
```

Our context
```
context = { current_user: current_user }
```

Changing this to a required keyword argument makes tests surface the bug. The code change fixes the tests.

### DIFF
--- a/lib/apollo-federation/schema.rb
+++ b/lib/apollo-federation/schema.rb
@@ -47,7 +47,7 @@ module ApolloFederation
         super
       end
 
-      def federation_sdl(context:)
+      def federation_sdl(context = nil)
         document_from_schema = FederatedDocumentFromSchemaDefinition.new(self, context: context)
         GraphQL::Language::Printer.new.print(document_from_schema.document)
       end
--- a/lib/apollo-federation/schema.rb
+++ b/lib/apollo-federation/schema.rb
@@ -47,7 +47,7 @@ module ApolloFederation
         super
       end
 
-      def federation_sdl(context = nil)
+      def federation_sdl(context: nil)
         document_from_schema = FederatedDocumentFromSchemaDefinition.new(self, context: context)
         GraphQL::Language::Printer.new.print(document_from_schema.document)
       end
--- a/lib/apollo-federation/schema.rb
+++ b/lib/apollo-federation/schema.rb
@@ -47,7 +47,7 @@ module ApolloFederation
         super
       end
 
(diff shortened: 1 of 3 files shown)


grxy:
@matthewjf Can you add a breaking test that reproduces this bug and show that your code fixes it?

lib/apollo-federation/schema.rb
```diff
@@ -47,7 +47,7 @@ def to_graphql
         super
       end
 
-      def federation_sdl(context: nil)
+      def federation_sdl(context:)
```
cmschuetz: Looks like I broke this when attempting to make this backwards compatible, context: https://github.com/Gusto/apollo-federation-ruby/pull/45#discussion_r370944253

I think if we default to `nil` the following way, we don't need to update `service_field.rb`

```suggestion
      def federation_sdl(context = nil)
```

matthewjf:
@grxy After a bit of digging around, it seems the intent of the original change was to support visibility checks when generating the sdl. There's no test coverage for that behavior, hence the bug. I'm not exactly sure how to test it. Can you point me in the right direction?

matthewjf:
@cmschuetz Let me know if those tests make sense. Just going off the visibility documentation as our apps don't use filters.

lib/apollo-federation/schema.rb
```diff
@@ -47,7 +47,7 @@ def to_graphql
         super
       end
 
-      def federation_sdl(context: nil)
```
grxy: Let's change this back to an optional keyword arg and update `ServiceField` to pass in `context` as a keyword arg.

PR #52, opened 2020-02-03
fix: Fix tracing with graphql-ruby interpreter

jturkel:
Attempting to use `ApolloFederation::Tracing` with the graphql-ruby [interpreter runtime](https://graphql-ruby.org/queries/interpreter.html) resulted in exceptions like this:

```
KeyError: key not found: :context
# ./lib/apollo-federation/tracing/tracer.rb:106:in `fetch'
# ./lib/apollo-federation/tracing/tracer.rb:106:in `execute_field'
# ./lib/apollo-federation/tracing/tracer.rb:50:in `trace'
# ...
```
This PR fixes that issue and resolves another issue with the `field_type` being set incorrectly in metrics recorded using the interpreter. The `spec/apollo-federation/tracing_spec.rb` diff looks kind of scary but I just moved all of the testing into a shared example group that could be reused with the legacy runtime and the interpreter runtime.

### DIFF
--- a/lib/apollo-federation/tracing/tracer.rb
+++ b/lib/apollo-federation/tracing/tracer.rb
@@ -103,7 +103,7 @@ module ApolloFederation
       # because we don't have the error `location` here.
       # rubocop:disable Metrics/CyclomaticComplexity, Metrics/PerceivedComplexity
       def self.execute_field(data, &block)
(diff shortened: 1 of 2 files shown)


srgoldman:
@noaelad any chance of getting this into an official release soon?

jturkel:
Looks like this ended up being fixed in #59.

PR #58, opened 2020-03-18
fix(lazy resolve): Handle problem with sync resolve

konalegi:
Hey!

Thanks for your work around this library.
When we are trying to implement authorization in our application we use exceptions and middleware, it works fine for everything except this gem, because the sync is being called [here](https://github.com/Gusto/apollo-federation-ruby/blob/master/lib/apollo-federation/entity_type_resolution_extension.rb#L7) which nullifies all data returned by `_entities` field even if the only one object is not authorized.

## Examples

What we get without this patch:
```
{
  "data": null,
  "errors": [
    {
      "message": "Not authorized (Talent)",
      "locations": [
        {
          "line": 2,
          "column": 3
        }
      ],
      "path": [
        "_entities"
      ],
      "extensions": {
        "code": "UNAUTHORIZED"
      }
    }
  ]
}
```
what we want to get: 
```
{
  "data": {
    "_entities": [
      {
        "id": "VjEtVGFsZW50LTEwMDAwMA",
        "status": "active",
        "resumeUrl": "/ConstructionWorker/resume/talent-1"
      },
      null
    ]
  },
  "errors": [
    {
      "message": "Not authorized (Talent)",
      "locations": [
        {
          "line": 2,
          "column": 3
        }
      ],
      "path": [
        "_entities",
        1
      ],
      "extensions": {
        "code": "UNAUTHORIZED"
      }
    }
  ]
}
```

### DIFF
--- a/lib/apollo-federation/entity_type_resolution_extension.rb
+++ b/lib/apollo-federation/entity_type_resolution_extension.rb
@@ -2,17 +2,15 @@
 
 class EntityTypeResolutionExtension < GraphQL::Schema::FieldExtension
   def after_resolve(value:, context:, **_rest)
-    synced_value =
-      value.map do |type, result|
-        [type, context.query.schema.sync_lazy(result)]
+    value.map do |type, result|
+      context.schema.after_lazy(result) do |resolved_value|
+        # TODO: This isn't 100% correct: if (for some reason) 2 different resolve_reference calls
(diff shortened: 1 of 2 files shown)


rylanc:
This looks good after my first pass. Do you mind fixing the merge conflicts and I'll take a final look?

konalegi:
@rylanc Done.

PR #59, opened 2020-03-20
fix: Make the tracer work with the new interpreter runtime

daemonsy:
## What's up? 

Context can be on the query on passed directly as denoted by this statement. By without passing a second argument to `fetch`, it raises so we never get to the `data.fetch(:query).context` branch.

### DIFF
--- a/lib/apollo-federation/tracing/tracer.rb
+++ b/lib/apollo-federation/tracing/tracer.rb
@@ -103,7 +103,7 @@ module ApolloFederation
       # because we don't have the error `location` here.
       # rubocop:disable Metrics/CyclomaticComplexity, Metrics/PerceivedComplexity
       def self.execute_field(data, &block)
-        context = data.fetch(:context) || data.fetch(:query).context
+        context = data.fetch(:context, nil) || data.fetch(:query).context
         return block.call unless context && context[:tracing_enabled]
 
         start_time_nanos = Process.clock_gettime(Process::CLOCK_MONOTONIC, :nanosecond)
--- a/lib/apollo-federation/tracing/tracer.rb
+++ b/lib/apollo-federation/tracing/tracer.rb
@@ -116,16 +116,18 @@ module ApolloFederation
 
         end_time_nanos = Process.clock_gettime(Process::CLOCK_MONOTONIC, :nanosecond)
 
-        # interpreter runtime
+        # legacy runtime
         if data.include?(:context)
           path = context.path
           field_name = context.field.graphql_name
           field_type = context.field.type.to_s
           parent_type = context.parent_type.graphql_name
-        else # legacy runtime
+        else # interpreter runtime
           path = data.fetch(:path)
-          field_name = data.fetch(:field).graphql_name
-          field_type = data.fetch(:field).type.unwrap.graphql_name
+          field = data.fetch(:field)
+          field_name = field.graphql_name
+          field_type = field.type.unwrap.graphql_name
+          field_type = field_type + '!' if field.type.non_null?
           parent_type = data.fetch(:owner).graphql_name
         end
 
--- a/lib/apollo-federation/tracing/tracer.rb
+++ b/lib/apollo-federation/tracing/tracer.rb
@@ -147,7 +147,7 @@ module ApolloFederation
       # Optional Step 3:
       # Overwrite the end times on the trace node if the resolver was lazy.
       def self.execute_field_lazy(data, &block)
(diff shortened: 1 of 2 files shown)


spec/apollo-federation/tracing_spec.rb
```diff
@@ -4,307 +4,320 @@
 require 'apollo-federation'
 
 RSpec.describe ApolloFederation::Tracing do
-  let(:base_schema) do
-    Class.new(GraphQL::Schema) do
-      use ApolloFederation::Tracing
+  RSpec.shared_examples 'a basic tracer' do
```
daemonsy: Large diff to the spec mostly due to indentation and an edit on one of the line. I'll highlight where the change is

spec/apollo-federation/tracing_spec.rb
```diff
@@ -4,307 +4,320 @@
...
+                  }, {
+                    response_name: 'name',
+                    type: 'String!',
+                    start_time: 9,
+                    end_time: 10,
+                    parent_type: 'Item',
+                    error: [{
+                      message: "Can't continue with this query",
+                      location: [{ line: 1, column: 15 }],
+                      json: %({"message":"Can't continue with this query","locations":[{"line":1,"column":15}],"path":["items",1,"name"]}),
```
daemonsy: I removed some whitespace here. Previously it had some whitespace added into it.

spec/apollo-federation/tracing_spec.rb
```diff
@@ -4,307 +4,319 @@
...
-      )
+  context 'with the new interpreter' do
+    let(:base_schema) do
+      Class.new(GraphQL::Schema) do
+        use ApolloFederation::Tracing
+        use GraphQL::Execution::Interpreter
+      end
     end
+
+    it_behaves_like 'a basic tracer'
```
daemonsy: New test case using the interpreter

lib/apollo-federation/tracing/tracer.rb
```diff
@@ -116,16 +116,18 @@ def self.execute_field(data, &block)
...
           field_type = context.field.type.to_s
           parent_type = context.parent_type.graphql_name
-        else # legacy runtime
+        else # interpreter runtime
           path = data.fetch(:path)
-          field_name = data.fetch(:field).graphql_name
-          field_type = data.fetch(:field).type.unwrap.graphql_name
+          field = data.fetch(:field)
+          field_name = field.graphql_name
+          field_type = field.type.unwrap.graphql_name
```
daemonsy: Good call to write a spec @grxy . This has always been broken by the test caught it. The `graphql_name` of the type doesn't include whether it's nullable, hence the need to check the field

lib/apollo-federation/tracing/tracer.rb
```diff
@@ -116,16 +116,18 @@ def self.execute_field(data, &block)
...
 
-        # interpreter runtime
+        # legacy runtime
         if data.include?(:context)
           path = context.path
           field_name = context.field.graphql_name
           field_type = context.field.type.to_s
           parent_type = context.parent_type.graphql_name
-        else # legacy runtime
+        else # interpreter runtime
```
daemonsy: I believe we got the legacy and new interpreter the wrong way round. I get into this branch

daemonsy:
Hey @noaelad :wave: this is a small fix before we can switch our own stuff to use the new interpreter.

spec/apollo-federation/tracing_spec.rb
grxy: Should we also test with the legacy runtime?

daemonsy: I added another test case. Good catch, I was under the impression that the examples run as a suite, but that doesn't make sense. 

Also, the `base_schema` variable wasn't been overwritten correctly. I removed it from the example so the actual test cases have to define it correctly.

lib/apollo-federation/tracing/tracer.rb
```diff
@@ -116,16 +116,17 @@ def self.execute_field(data, &block)
...
           path = data.fetch(:path)
-          field_name = data.fetch(:field).graphql_name
-          field_type = data.fetch(:field).type.unwrap.graphql_name
+          field = data.fetch(:field)
+          field_name = field.graphql_name
+          field_type = field.type.to_type_signature
```
daemonsy: This seems like it's what we want. The other hacky way was leaky

PR #62, opened 2020-04-06
fix: Fix spec to account for breaking change to graphql-ruby 1.10.0

kryptonat:
Account for [breaking change](https://github.com/rmosolgo/graphql-ruby/blob/master/CHANGELOG.md#1100-20-jan-2020) to `graphql-ruby 1.10.0` in spec

> Class-based schemas using the interpreter must add use GraphQL::Analysis::AST to their schema (and update their custom analyzers, see https://graphql-ruby.org/queries/ast_analysis.html) #2363


rylanc:
In the future, PRs that only have test changes/fixes should be marked as a chore. Changes like this don't affect the released code and therefore don't need a release.

Issue #63, opened 2020-04-21
Cannot use new interpreter with apollo-federation-ruby


tjwallace:
Using the [new `graphql-ruby` interpreter](https://graphql-ruby.org/queries/interpreter.html) with `apollo-federation-ruby` does not work.

* Query: `{ _service { sdl } }`
* Error: `Field '_service' doesn't exist on type 'Query'`

### Schema
```ruby
class AppSchema < GraphQL::Schema
  include ApolloFederation::Schema

  use GraphQL::Execution::Interpreter
  use GraphQL::Analysis::AST

  # ...
end
```

### Versions
* `graphql-ruby`: `1.10.7`
* `apollo-federation-ruby`: `1.0.4`

**Note:** The new interpreter + `apollo-federation-ruby` did work with `graphql-ruby` `1.9.19`.

bilby91:
Having the same issue with GraphQL 1.10

bilby91:
I've been doing some experimentation with adding support for the new interpreter. I have it working but the changes will probably require to change the approach on how the federation is setup.

I can probably send a MR to discuss this. I also found that the current implementation of overriding the root query breaks other dependencies like `graphql-guard`.

bilby91:
I'm trying to see if we can use/build an API that would let us hook the different pieces that are needed for federation here https://github.com/rmosolgo/graphql-ruby/issues/2928

PR #64, opened 2020-05-11
WIP: Add support for new interpreter

### LINKED ISSUE #48: Support for graphql 1.10+
Hey there!

I've just started playing with this gem, so far so good, tks for getting this released :beer:  While I was experimenting with things, I tried to update to the recently released [graphql 1.10.0](https://github.com/rmosolgo/graphql-ruby/blob/master/CHANGELOG.md#1100-20-jan-2020) but found out that this gem is not compatible with it.

Was there a specific reason for locking the dep to 1.9.x? I haven't gone through the sources of this project yet and there are some breaking changes on 1.10.0 (and it is also supposed to yield [faster boot times](https://github.com/rmosolgo/graphql-ruby/pull/2363#issuecomment-562191562), which is why we were considering upgrading it :zap: )

bilby91:
This change is a work in progress to get federation working with the new GraphQL ruby interpreter.

There are several breaking changes and I still think that a different approach should be taken in order to enrich the root query.

In order for using this version of the code, the following changes need to be done:

In the schema class:

```
  query(Types::QueryType)
  use(GraphQL::Execution::Interpreter)
  use(GraphQL::Analysis::AST)
  use(ApolloFederation::Schema) # <---- If this is done before defining the query it won't work.
end
```

This was not needed before, now it's needed because we don't automagically add the fields to the root query.
```
class Types::Query < GraphQL::Schema::Object
    include ApolloFederation::ServiceField
    include ApolloFederation::EntitiesField

    # Define the federation _entities field
    define_entities_field(ApolloFederation::Entity)
end
```

Beside this two changes there are two other caveats. 

If the root query is not defined with the included modules, nothing from federation will be added. 

If the Entity type has zero candidate for the possible_types, an error would be raised stating that the Entity union type has zero possible types.

Some interesting feedback from the GraphQL ruby author can be found here: https://github.com/rmosolgo/graphql-ruby/issues/2928

Closes #48 #63


bilby91:
Found another interesting problem with the `_service` field definition. Since the `_service` was not added when the module got included, the owner of the `_service` field was the `ApolloFederation::ServiceField` module instead of the root query. I moved to a similar approach like the one that is used for entities. Made a sugar module that groups both _serivce and _entities.

bilby91:
Found another issue with the previous approach. 

Since the `field(:_service, Service, null: false)` call is done directly when the `ApolloFederation::ServiceField` is included, the owner of the field is assigned to the module instead of the root `Query`. This can problems with authorization gems like graphql-guard. Moving to the approach of explicitly including the module and calling a method to define the field solves the problem.

bilby91:
@rylanc Thanks for the feedback! I'm looking forward to find some common ground and get this moving 💪🏻

I'm on the same page in terms of making this transparent for the user. The `define_X_fields` API can be hidden and called when including the module, that shouldn't be a problem.

In regards to making non breaking changes, I think that is doable but would probably require having two different mechanisms for "setting up" the library.

I've been reading the different specs and I think that the current implementation leads to having an API that could not be compliant with the Apollo Federation spec. 

This are the things that are currently allowed and I think that they shouldn't.

If the user does not define any type that is a possible type of the `_Entity` type, the current implementation will not add the `_entities` field. This would break the Apollo Federation specification that states that the `_entities` field is not optional. I think that if this happens we should be raising a clear error to the user stating that something is wrong.

I like the idea of using the `use` API instead of doing the `include` that we are currently doing. The only caveat with this is the order. If we go with the `use` API, we need to make sure that is really clear in the documentation that the `use` call needs to happen after all the types are added in the schema (basically calling `mutation`, `query`, `subscriptions` and `orphan_types`). I really don't love this but I haven't found a better way to solve this problem without making changes (that would be accepted) in the GraphQL gem.

Looking forward on your thoughts!

rylanc:
@bilby91 I was able to dedicate some more time to this today and I came up with a solution that seems to work without making any breaking changes: https://github.com/Gusto/apollo-federation-ruby/pull/65. Please take a look and let me know what you think.

At some point, I would like to move to the `use` approach, but I'd like to get a non-breaking version out first and compile a list of breaking changes I'd like to make so I can make the migration easier for users.

Where in the specification did you see that bit about leaving the `_entities` field in? From my research, [it looks like it should be removed if there are no "entities"](https://www.apollographql.com/docs/apollo-server/federation/federation-spec/#resolve-requests-for-entities):

> To implement the `_Entity` union, each type annotated with `@key` should be added to the _Entity union. **If no types are annotated with the key directive, then the `_Entity` union and `Query._entities` field should be removed from the schema.**

Again, thanks for digging into this. The issue you filed on the `graphql-ruby` repo was really helpful.

bilby91:
@rylanc Thanks so much for the feedback and  the PR. I sent some suggestions.

I clearly missed that statement on the spec. My understanding was that the `_entities` field was not optional at all and that seems to not be true.

Thanks again!

bilby91:
Closing in favor of #65

PR #65, opened 2020-05-15
feat: Add support for interpreter runtime

rylanc:
Fixes https://github.com/Gusto/apollo-federation-ruby/issues/63

### DIFF
--- a/lib/apollo-federation/schema.rb
+++ b/lib/apollo-federation/schema.rb
@@ -5,45 +5,67 @@ require 'apollo-federation/service_field'
 require 'apollo-federation/entity'
 require 'apollo-federation/federated_document_from_schema_definition.rb'
 
+# Random ideas:
+#  - Create a temp schema inside .query and use that to extract the types
+#  - Modify the query object after the fact? I doubt it would work...
+
 module ApolloFederation
   module Schema
     def self.included(klass)
+      # # TODO: Maybe make this a "use" interface?
       klass.extend(ClassMethods)
     end
 
     module ClassMethods
-      def to_graphql
-        orig_defn = super
+      # TODO: Hmmm. Is there a way to clear everything in the schema?
+      def query(new_query_object = nil)
+        if new_query_object
+          super
+        else
+          @federation_query_object ||= begin
+            # TODO: Better organize this method
+            # Walk through all of the types and determine which ones are "entities"
+            possible_entities = types.values.select do |type|
+              !type.introspection? && type.include?(ApolloFederation::Object) &&
+                type.federation_directives&.any? { |directive| directive[:name] == 'key' }
+                # type.metadata[:federation_directives]&.any? { |directive| directive[:name] == 'key' }
+            end
 
-        possible_entities = orig_defn.types.values.select do |type|
-          !type.introspection? && !type.default_scalar? && type.is_a?(GraphQL::ObjectType) &&
-            type.metadata[:federation_directives]&.any? { |directive| directive[:name] == 'key' }
-        end
+            # Build the new query object with the '_service' field
+            federation_query_object = federation_query(super)
 
-        @query_object = federation_query
+            # If there are any "entities", define the Entity union and and the Query._entities field
+            if !possible_entities.empty?
+              entity_type = Class.new(Entity) do
+                possible_types(*possible_entities)
+              end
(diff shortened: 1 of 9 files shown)


bilby91:
Will give it a try on my repo and check if tests pass!

lib/apollo-federation/schema.rb
```diff
@@ -12,49 +12,74 @@ def self.included(klass)
...
         elsif Gem::Version.new(GraphQL::VERSION) >= Gem::Version.new('1.10.0')
-          base = query
+          base = query_obj
         else
-          base = query.metadata[:type_class]
+          # TODO: Will this ever happen?
+          base = query_obj.metadata[:type_class]
         end
 
         Class.new(base) do
```
bilby91, line 74: Wrapping the query type in another class can cause issues with other gems like this one https://github.com/exAspArk/graphql-guard#policy-object

What if we do it with class_eval ? My test suite broke because of this change.

```
base.class_eval do
  include EntitiesField
  include ServiceField
```

lib/apollo-federation/schema.rb
```diff
@@ -12,49 +12,74 @@ def self.included(klass)
...
+          # TODO: Maybe the name should inherit from the original Query name
+          # Or MAYBE we should just modify the original class?
           graphql_name 'Query'
```
bilby91, line 77: What do you think of `graphql_name base. graphql_name || 'Query'`  here ?

lib/apollo-federation/schema.rb
```diff
@@ -12,49 +12,74 @@ def self.included(klass)
...
-        @query_object = federation_query
+            # If there are any "entities", define the Entity union and and the Query._entities field
+            possible_entities = schema_entities
+            if !possible_entities.empty?
+              entity_type = Class.new(Entity) do
+                possible_types(*possible_entities)
+              end
+              # TODO: Should/can we encapsulate all of this inside the module? What's the best/most
+              # Ruby way to split this out?
+              @federation_query_object.define_entities_field(entity_type)
```
bilby91, line 31: Defining the `_entities` field like this sets the correct parent to the type. This is not true with the `_service` field. Can we refactor how the `_service` field is included an do the same as `_entities`.

Basically, in this method we would call at some point:

```
@federation_query_object.define_service_field
```

bilby91:
@rylanc This looks pretty good! There are some small issues that I faced with this implementation but I think we can fix them really easy without generating breaking changes.

Let me know what you think! I'm happy to make those change on top of this PR if needed.

lib/apollo-federation/schema.rb
rylanc: Interesting. Is this something that's currently broken in v1.0.4? If so, I'd probably rather fix it in another PR/issue.

lib/apollo-federation/schema.rb
rylanc: Yeah. Good call

lib/apollo-federation/schema.rb
rylanc: Hmmm. Can you clarify what you mean by "parent"? Is the missing parent causing issues? Also, (like above) is this something that was already broken in v1.0.4?

rylanc:
@bilby91 Thanks for taking a look. If any of the issues you mentioned were broken before this PR (e.g. in v1.0.4), I'd rather wait and address them in separate PRs (with help from you, if you're up for it).

lib/apollo-federation/schema.rb
bilby91: Yes. Let's do it in a different PR!

lib/apollo-federation/schema.rb
bilby91: Sorry, I should have said `owner`, not `parent`. If I recall correctly this `self` is evaluated as the module and not the query type.

https://github.com/rmosolgo/graphql-ruby/blob/master/lib/graphql/schema/member/has_fields.rb#L12

bilby91:
@rylanc I can send those changes as different PRs.

Once this change is merged I'll submit the other!

bilby91:
Is there any ETA for merging this change :) ?

rylanc:
Sorry for the delay. I added a CI matrix build to test compatibility with `graphql-ruby` 1.9 and it looks like it's broken ☹️ . I'm going to try to fix it today

PR #66, opened 2020-05-20
ci: Add matrix builds to test graphql 1.9 and 1.10

rylanc:


kryptonat:
Looks like `ci/circleci: integration-tests` and `ci/circleci: ruby-test` commit status is stuck on pending

rylanc:
It's because they're required in the Github settings, but they don't exist in this branch (they've been replaced by the ones with the explicit gemfiles). Once this branch lands, I was planning on changing those settings to match our new CI config.

PR #70, opened 2020-05-28
fix: Fix _service field type owner

bilby91:
The field was previously defined in the module making the
ApolloFederation::ServiceField the owner of the type instead of the
root Query.

### DIFF
--- a/graphql-1.9.gemfile.lock
+++ b/graphql-1.9.gemfile.lock
@@ -1,7 +1,7 @@
 PATH
   remote: .
   specs:
-    apollo-federation (1.1.0)
+    apollo-federation (1.1.1)
       google-protobuf (~> 3.7)
       graphql (>= 1.9.8)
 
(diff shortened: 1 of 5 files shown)


bilby91:
The failures for 1.9 seem unrelated to the changes. Is there anything that I might be missing ?

rylanc:
Thanks for doing this. Just so I understand: is this issue breaking any functionality? Or is it just for correctness?

bilby91:
Thanks for looking into it @rylanc . Tests should be fixed now.

This is currently breaking my usage of [graphql-guard](https://github.com/exAspArk/graphql-guard).

I just run my test suite without this fix and I can't protect the `_service` field (https://github.com/exAspArk/graphql-guard/blob/fb42e72112c51cde2380c0a62d31fb3d63515d45/lib/graphql/guard.rb#L67).

bilby91:
Anything else I do @rylanc to get this moving ? 

Thanks!

bilby91:
Ping @rylanc :)

PR #75, opened 2020-06-05
fix: Add types with fields that @requires to _Entity union

barthez:
Types that have fields that uses `@requires` directive needs to be part of `_Entity` union and be resolvable by `_entities` query field.

It is useful when one schema is extending types that are kinda of "singletons" like `viewer`, eg:
```
# Accounts service
type Query {
  viewer: Viewer!
}

type Viewer {
  my_products: [Product!]!
}

# Reviews service
type Viewer @extend {
  my_products: [Product!]! @external
  my_product_reviews: [Review!]! @requires fields: 'my_products { id }'
}
```
The query to gateway:
```
query {
  viewer {
     my_product_reviews {
       body
     }
  }
}
```
would get Viewer form accounts service and try to resolve `my_product_reviews` from reviews service using _entities query. Because there is no `@key` directive on viewer it is not part of `_Entity` union and `_entities` query would fail. 

The change makes the gem to add types with fields with `@requires` directive to be part of `_Entities` union.

### DIFF
--- a/lib/apollo-federation/schema.rb
+++ b/lib/apollo-federation/schema.rb
@@ -45,6 +45,10 @@ module ApolloFederation
         klass.define_service_field
         klass
       end
+
+      def includes_directive?(directives, name)
+        directives&.any? { |directive| directive[:name] == name }
+      end
     end
 
     # TODO: Remove these once we drop support for graphql 1.9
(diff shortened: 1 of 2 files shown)


barthez:
Hey @rylanc, can I get your insights on this?

barthez:
@rylanc any chance you could take a peek on this?

barthez:
Hey @rylanc, would you mind taking a look at the PR? Is it suitable to be merge? Should I change something?

daemonsy:
Closing this, see discussion in #192

PR #83, opened 2020-06-10
ci: Use appraisal for matrix builds

rylanc:


Gemfile.lock
```diff
@@ -34,7 +38,7 @@ GEM
     crass (1.0.6)
     diff-lcs (1.3)
     erubi (1.9.0)
-    google-protobuf (3.12.2)
+    google-protobuf (3.12.2-universal-darwin)
```
avalanche123, line 41: seems osx specific, is that intentional?

rylanc: It wasn't intentional. I was thinking the same thing, but it didn't seem to break anything in CI. I'll revert it just in case.

PR #94, opened 2020-08-07
fix: Inherit federation directives

jturkel:
This PR adds inheritance of federation directives for consistency with graphql-ruby's native directives and other parts of their API like `field`, `field_class`, etc. Code like this is now supported:

```ruby
class RemotelyReferenceableObject < SalsifyGraphQL::Server::BaseObject
  key fields: 'id'
  field :id, ID, null: false
end

class BlogType < RemotelyReferenceableObject
  field :text, String, null: false
end

class BaseSchema <  GraphQL::Schema
  federation version: '2.0'
end

class MySchema < BaseSchema
  # ...
end
```

### DIFF
--- a/lib/apollo-federation/has_directives.rb
+++ b/lib/apollo-federation/has_directives.rb
@@ -7,8 +7,8 @@ module ApolloFederation
     end
 
     def federation_directives
-      if is_a?(Class) && superclass.respond_to?(:federation_directives)
-        own_federation_directives + superclass.federation_directives
+      if is_a?(Class)
+        own_federation_directives + find_inherited_value(:federation_directives, [])
       else
         own_federation_directives
       end
--- a/lib/apollo-federation/has_directives.rb
+++ b/lib/apollo-federation/has_directives.rb
@@ -2,11 +2,20 @@
 
 module ApolloFederation
   module HasDirectives
-    attr_reader :federation_directives
-
     def add_directive(name:, arguments: nil)
-      @federation_directives ||= []
-      @federation_directives << { name: name, arguments: arguments }
+      own_federation_directives << { name: name, arguments: arguments }
+    end
+
+    def federation_directives
+      if is_a?(Class) && superclass.respond_to?(:federation_directives)
+        own_federation_directives + superclass.federation_directives
+      else
+        own_federation_directives
+      end
+    end
+
+    def own_federation_directives
+      @own_federation_directives ||= []
     end
   end
 end
--- a/lib/apollo-federation/schema.rb
+++ b/lib/apollo-federation/schema.rb
@@ -22,7 +22,7 @@ module ApolloFederation
       end
 
       def federation_version
-        @federation_version || '1.0'
+        @federation_version || (superclass.respond_to?(:federation_version) && superclass.federation_version) || '1.0'
       end
 
       def federation_2?
@@ -38,7 +38,11 @@ module ApolloFederation
       end
 
       def link_namespace
-        @link[:as]
+        if @link
+          @link[:as]
+        elsif superclass.respond_to?(:link_namespace)
+          superclass.link_namespace
+        end
       end
 
       def query(new_query_object = nil)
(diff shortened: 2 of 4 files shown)


jturkel:
Note #184 is an alternate way to solve the problem of directives not being inherited.

jturkel:
@grxy - Any chance you could review this? We'd love to remove our monkey patches to workaround this problem.

lib/apollo-federation/has_directives.rb
```diff
@@ -2,11 +2,20 @@
...
   module HasDirectives
-    attr_reader :federation_directives
-
     def add_directive(name:, arguments: nil)
-      @federation_directives ||= []
-      @federation_directives << { name: name, arguments: arguments }
+      own_federation_directives << { name: name, arguments: arguments }
+    end
+
+    def federation_directives
```
grxy, line 9: Would you mind updating this implementation to use `GraphQL::Schema::FindInheritedValue`?

lib/apollo-federation/schema.rb
```diff
@@ -22,7 +22,7 @@ def federation(version: '1.0', link: {})
       end
 
       def federation_version
-        @federation_version || '1.0'
+        @federation_version || (superclass.respond_to?(:federation_version) && superclass.federation_version) || '1.0'
```
grxy, line 25: Would you mind updating this implementation to use `GraphQL::Schema::FindInheritedValue`?

lib/apollo-federation/schema.rb
```diff
@@ -38,7 +38,11 @@ def federation_sdl(context: nil)
       end
 
       def link_namespace
-        @link[:as]
+        if @link
```
grxy, line 41: Would you mind updating this implementation to use `GraphQL::Schema::FindInheritedValue`?

spec/apollo-federation/service_field_v2_spec.rb
```diff
@@ -1539,6 +1539,50 @@ def self.visible?(context)
...
+          type Product @federation__key(fields: "id") {
+            id: String!
+          }
+        GRAPHQL
+      )
+    end
+
+    it 'returns SDL that inherits schema directives' do
+      new_base_schema = Class.new(base_schema) do
+        federation version: '2.0', link: { as: 'fed2' }
```
grxy, line 1572: Since you are adding a custom link namespace, can you add an example to the test so we can see that directives are being prefixed with `fed2__` instead of the default?

spec/apollo-federation/service_field_v1_spec.rb
```diff
@@ -455,6 +455,30 @@ def self.visible?(context)
       )
     end
 
+    it 'returns SDL that inherits object directives' do
+      base_object_with_id = Class.new(base_object) do
+        key fields: 'id'
+
+        field :id, String, null: false
```
grxy, line 462: nit: Can you make this a field of type ID?

spec/apollo-federation/service_field_v1_spec.rb
```diff
@@ -455,6 +455,30 @@ def self.visible?(context)
...
+      end
+
+      expect(execute_sdl(schema)).to match_sdl(
+        <<~GRAPHQL,
+          type Product @key(fields: "id") {
+            id: String!
+          }
+        GRAPHQL
+      )
+    end
```
grxy, line 480: Do you want to add a schema inheritance test here as well?

grxy (changes requested):
Thanks for rebasing the PR. Have a few requests/suggestions.

spec/apollo-federation/service_field_v1_spec.rb
jturkel: The [v1](https://www.apollographql.com/docs/federation/v1/federation-spec/#federation-schema-specification) federation spec doesn't support any schema level directives so I don't think there's anything to test here.

grxy (approved):

jturkel:
Thanks for the review @grxy!

grxy:
@jturkel No prob! I think the other PRs merged w/ the wrong commit message, so all of those PRs will probably be going out in a patch version instead of a minor version as intended. Whoops. 😅

PR #95, opened 2020-08-28
fix: Fix context passed to GraphQL::Schema.federation_sdl

jturkel:
This fixes a problem where `ApolloFederation::ServiceField#_service` passed a `context` of type `GraphQL::Query::Context` rather than `Hash` which violates the contract of [GraphQL::Language::DocumentFromSchemaDefinition#initialize](https://github.com/rmosolgo/graphql-ruby/blob/master/lib/graphql/language/document_from_schema_definition.rb#L9). The API of `GraphQL::Query::Context` and `Hash` are similar enough that this didn't cause any problems until graphql-ruby 1.9.17 with the introduction of [context scoping](https://github.com/rmosolgo/graphql-ruby/pull/2634). After that change, calling `to_h`, `fetch` or `dig` on the context from anywhere in the resolution of `ApolloFederation::ServiceField#_service` (e.g. schema visibility checks) would result in a ` NoMethodError: undefined method `merge' for #<Query::Context ...>`.

### DIFF
--- a/Appraisals
+++ b/Appraisals
@@ -1,11 +1,11 @@
 # frozen_string_literal: true
 
 appraise 'graphql-1.9' do
-  gem 'graphql', '1.9.8'
+  gem 'graphql', '1.9.21'
 end
 
 appraise 'graphql-1.10' do
-  gem 'graphql', '1.10.10'
+  gem 'graphql', '1.10.14'
 end
 
 appraise 'graphql-1.11' do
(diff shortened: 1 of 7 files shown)


rylanc (changes requested):
This looks great! Could you rebase/merge with master so the graphql-1.11 CI checks are run?

jturkel:
@rylanc - I rebased on top of master and the tests are passing.

jturkel:
@rylanc - Looks like another merge conflict snuck in so I rebased the PR.

rylanc (approved):

PR #102, opened 2020-09-22
fix(tracing): properly handle parsing and validation errors

rylanc:


lib/apollo-federation/tracing/tracer.rb
```diff
@@ -35,15 +35,15 @@ module ApolloFederation
...
       EXECUTE_QUERY_LAZY = 'execute_query_lazy'
       EXECUTE_FIELD = 'execute_field'
       EXECUTE_FIELD_LAZY = 'execute_field_lazy'
 
       def self.trace(key, data, &block)
         case key
-        when EXECUTE_QUERY
-          execute_query(data, &block)
+        when EXECUTE_MULTIPLEX
+          execute_multiplex(data, &block)
```
noaelad, line 46: What is a multiplex? [The docs](https://graphql-ruby.org/api-doc/1.9.6/GraphQL/Tracing) are really sparse

noaelad: NVM [figured it out](https://graphql-ruby.org/queries/multiplex.html)

lib/apollo-federation/tracing.rb
```diff
@@ -23,37 +15,10 @@ def should_add_traces(headers)
       headers && headers['apollo-federation-include-trace'] == KEY.to_s
     end
 
-    def attach_trace_to_result(result)
-      return result unless result.context[:tracing_enabled]
-
-      trace = result.context.namespace(KEY)
-      # TODO: If there are errors during query validation, that could also cause a missing
-      # start_time
```
noaelad, line 31: Is this issue addressed anywhere?

spec/apollo-federation/tracing_spec.rb
```diff
@@ -511,6 +507,58 @@ def items
...
+                      typeName: 'Query',
+                      fieldName: 'nonExistant',
+                    },
+                  }.to_json,
+                }],
+              },
+            ),
+          )
+        end
+      end
```
noaelad, line 561: 👍 💯

.circleci/config.yml
```diff
@@ -108,9 +108,6 @@ workflows:
             parameters:
               appraisal: ["graphql-1.9", "graphql-1.10", "graphql-1.11"]
       - release:
-          filters:
-            branches:
-              only: master
```
noaelad, line 113: I think we still want this...?

noaelad (changes requested):
This looks great! Left just a couple of questions.

lib/apollo-federation/tracing.rb
rylanc: Yeah, thats the main thing this is fixing. The start time is now recorded at the beginning of the `execute_multiplex` event, which will always happen before any parsing or validation. Also, now that the trace is added to the result automatically, we don't need to check if the tracer was properly installed (since the "add traces to result" code won't even run if the tracer isn't installed).

.circleci/config.yml
rylanc: I removed this because we don't really need it and it allows us to publish prerelease versions. If we always run it, it just lets semantic-release decide if there should be a release. For the main branch (`master`), it will publish a regular release; for a [prerelease branch](https://semantic-release.gitbook.io/semantic-release/usage/workflow-configuration#pre-release-branches) (e.g. `beta`), it will publish a prerelease and for all others it just won't publish. That's how I was able to publish a beta and test it.

noaelad (approved):
LGTM!

PR #105, opened 2020-10-02
feat: Snake case field references

### LINKED ISSUE #36: Consider snake_case for fields
While implementing the federation spec for a few types, I got tripped up when referencing fields in the `key` helper and `resolve_reference`.  Since all of the fields in `graphql-ruby` use snake case, I sometimes forget that the helpers for federation use the actual camelCased field names.

When authoring this library, was snake_case considered, and if so were there any reasons to not use it?  I feel as though the mix of snake & camel case may be slightly confusing, what are your thoughts?

jturkel:
This is an awesome library but one of the things I initially found confusing was the naming differences between graphql-ruby fields (snake case symbols) and apollo-federation-ruby field references (lower camelcase) e.g.

```ruby
class Order < BaseObject
  field :zip_code, String, null: true, external: true
  field :weight, Int, null: true, external: true
  field :shipping_estimate, Int, null: true, requires: { fields: 'zipCode weight' }
end
```

This PR updates field set arguments to be either strings which follow the Apollo syntax or arrays, hashes and snake case symbols that follow the graphql-ruby conventions e.g.

```ruby
class Order < BaseObject
  field :zip_code, String, null: true, external: true
  field :weight, Int, null: true, external: true
  field :shipping_estimate, Int, null: true, requires: { fields: [:zip_code, :weight] }
end
```
Supporting the existing string syntax makes the change backwards compatible and it also provides an escape hatch for advanced use cases like directives/arguments in field selections or not wanting camelization.
 
Fixes #36

### DIFF
--- a/example/inventory.rb
+++ b/example/inventory.rb
@@ -18,13 +18,13 @@ INVENTORY = [
 
 class Product < BaseObject
   extend_type
-  key fields: 'upc'
+  key fields: :upc
 
   field :upc, String, null: false, external: true
   field :weight, Int, null: true, external: true
   field :price, Int, null: true, external: true
   field :in_stock, Boolean, null: true
-  field :shipping_estimate, Int, null: true, requires: { fields: 'price weight' }
+  field :shipping_estimate, Int, null: true, requires: { fields: %i[price weight] }
 
   def self.resolve_reference(reference, _context)
     reference.merge(INVENTORY.find { |product| product[:upc] == reference[:upc] })
--- a/spec/apollo-federation/field_set_serializer_spec.rb
+++ b/spec/apollo-federation/field_set_serializer_spec.rb
@@ -1,0 +1,36 @@
+# frozen_string_literal: true
+
+require 'spec_helper'
+require 'apollo-federation/field_set_serializer'
+
+RSpec.describe ApolloFederation::FieldSetSerializer do
+  it 'serializes symbols' do
+    expect(
+      described_class.serialize(:id),
+    ).to eql('id')
+  end
+
+  it 'serializes snake case symbols as lower camelcase' do
+    expect(
+      described_class.serialize(:product_id),
+    ).to eql('productId')
+  end
+
+  it 'serializes arrays of symbols' do
+    expect(
+      described_class.serialize(%i[qualifier id]),
+    ).to eql('qualifier id')
+  end
+
+  it 'serializes nested selections' do
+    expect(
+      described_class.serialize([:id, organization: :id]),
+    ).to eql('id organization { id }')
+  end
+
+  it 'serializes strings without modification' do
+    expect(
+      described_class.serialize('product_id'),
+    ).to eql('product_id')
+  end
+end
--- a/spec/apollo-federation/service_field_spec.rb
+++ b/spec/apollo-federation/service_field_spec.rb
@@ -217,7 +217,7 @@ RSpec.describe ApolloFederation::ServiceField do
       it 'returns valid SDL for @key directives' do
         product = Class.new(base_object) do
           graphql_name 'Product'
-          key fields: 'upc'
+          key fields: :upc
 
           field :upc, String, null: false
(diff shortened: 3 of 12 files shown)


geshwho:
Hi @jturkel. I want to first apologize for the lack of response on this PR. I am a Gusto employee, and I, along with a colleague, have recently volunteered to help maintain this gem and we're slowly getting organized.

Back to the PR, I do think this would be a good feature to add; this was one of the things that confused me at first too. Hoping to review this by the end of the week. In the meantime, the branch has fallen out of date with `main`. Feel free to update it, if you have time. (If not, we can take over and get it across the finish line 🙂)

jturkel:
Hi @geshwho - I'm excited to see this project moving again 🎉  The change has been rebased.

example/inventory.rb
```diff
@@ -18,13 +18,13 @@
...
   extend_type
-  key fields: 'upc'
+  key fields: :upc
 
   field :upc, String, null: false, external: true
   field :weight, Int, null: true, external: true
   field :price, Int, null: true, external: true
   field :in_stock, Boolean, null: true
-  field :shipping_estimate, Int, null: true, requires: { fields: 'price weight' }
+  field :shipping_estimate, Int, null: true, requires: { fields: %i[price weight] }
```
geshwho, line 27: I'd prefer not to use the `%` notation to be consistent with the documentation in some other places.

```suggestion
  field :shipping_estimate, Int, null: true, requires: { fields: [:price :weight] }
```

spec/apollo-federation/field_set_serializer_spec.rb
```diff
@@ -0,0 +1,36 @@
...
+
+  it 'serializes snake case symbols as lower camelcase' do
+    expect(
+      described_class.serialize(:product_id),
+    ).to eql('productId')
+  end
+
+  it 'serializes arrays of symbols' do
+    expect(
+      described_class.serialize(%i[qualifier id]),
```
geshwho, line 21: Can we change this one too?

spec/apollo-federation/service_field_spec.rb
```diff
@@ -358,12 +358,12 @@ def execute_sdl(schema)
...
         graphql_name 'Product'
         extend_type
-        key fields: 'upc'
+        key fields: :upc
 
         field :upc, String, null: false, external: true
         field :weight, Integer, null: true, external: true
         field :price, Integer, null: true, external: true
-        field :shipping_estimate, Integer, null: true, requires: { fields: 'price weight' }
+        field :shipping_estimate, Integer, null: true, requires: { fields: %i[price weight] }
```
geshwho, line 366: Last one 🙂

geshwho (changes requested):
Requesting some really minor changes. After that, this LGTM 🎉

geshwho:
It's good that this PR is backwards compatible, but it is a bit confusing that strings and symbols do not have the same behavior. I think ultimately, we may want to make a breaking change that camelizes strings as well, but offers some configuration option to turn off camelization altogether. This would be more consistent with the way that graphql-ruby works as well.

I'll make an issue to track this. @jturkel let us know if you have any interest in helping with this.

example/inventory.rb
jturkel: I'm not a fan of readability of the `%i` notation either 🙂  This project uses the default Rubocop configuration for the `Style/SymbolArray` cop which requires it though. I'm happy to change the Rubocop configuration if you want.

geshwho: Ah.. Sorry, I did not notice the Rubocop config. I'll approve this change as is, and we can separately discuss what we want to do about styles.

geshwho (approved):

jturkel:
Consistent treatment of symbols and strings and an option to disable camelization sounds like a good idea. We (Salsify) aren't actually using Apollo Federation so I've got limited bandwidth to help out on it though.

PR #107, opened 2020-10-14
fix: Support inline variables for _entities

cmschuetz:
**Summary**
- Updates Any.coerce_input to handle the case when `value` is an instance of `GraphQL::Language::Nodes::InputObject`
- Newer releases of GraphQL automatically convert inline query variables to input objects

While it's highly unlikely that Apollo Gateway will send queries with inline variables, when unit testing & manually sending `_entitiy` queries directly to federation services, we often will use inline variables.


cmschuetz:
Hey @rylanc, any chance you could give this a look?

caomania:
Would like to +1 merging this fix.

daemonsy:
Sorry for taking so long to get to this PR. 

This makes sense, we should be supporting inlined variables. Let me see if I can revive this PR.

cmschuetz:
I'm gonna close this one team, i'm so far removed from this stuff these days.  Hope you found a path forward here

PR #110, opened 2020-10-22
fix(docs): update readme for adding Apollo Federation tracing

matthewjf:
I made a request to my rails server running locally and set the header `apollo-federation-include-trace`

```ruby
headers
=> {"X-Frame-Options"=>"SAMEORIGIN",
 "X-XSS-Protection"=>"1; mode=block",
 "X-Content-Type-Options"=>"nosniff",
 "X-Download-Options"=>"noopen",
 "X-Permitted-Cross-Domain-Policies"=>"none",
 "Referrer-Policy"=>"strict-origin-when-cross-origin"}

headers['apollo-federation-include-trace']
=> nil

request.headers['apollo-federation-include-trace']
=> "hello world"
```

Tracing shows up correctly when we use `request.headers` instead. A little confusing, but `headers` in a rails controller is a hash, while `request.headers` is an instance of `ActionDispatch::Http::Headers`.


brettfishman:
We ran into this same issue. ➕1

daemonsy:
Thanks for this @matthewjf. Using `request.headers` :+1:

PR #114, opened 2021-02-01
fix(tracing): headers recognition in Rails apps

siemakuba:
Inside a Ruby on Rails application custom request headers are transformed by uppercasing and prepending the `HTTP_` prefix before it reaches the application controller. This causes the `should_add_traces` method to not recognize the incoming `apollo-federation-include-trace` header properly. 

This is a simple fix that would properly return `true` for both the original and transformed header.

Rails API reference: https://github.com/rails/rails/blob/5f3ff60084ab5d5921ca3499814e4697f8350ee7/actionpack/lib/action_dispatch/http/headers.rb#L126


siemakuba:
Turns out it was my mistake during testing. Sorry for the confusion.

PR #135, opened 2021-08-03
fix: allow the federation schema to be a subclass of the standard schema

mziwisky:
i.e., allow something like:

      query MyQuery
    end

    class MyFederationSchema < MySchema
    end

this pattern is useful if you want to host both a federation subgraph
_and_ a "first-class" data graph with the same schema minus the
federation extensions.

this didn't work before this change, because `MyFederationSchema` never
gets `query MyQuery` called on it, so it doesn't know about the
provided Query object.

the changes in the specs from subclassing `base_schema` to re-opening
and modifying it aren't necessary, but i think better reflect real-world
usage -- people probably don't subclass an empty base schema and then
add `query`, `mutation`, etc. to it, they just add those things to the
schema class itself.

### DIFF
--- a/lib/apollo-federation/schema.rb
+++ b/lib/apollo-federation/schema.rb
@@ -73,7 +73,7 @@ module ApolloFederation
           @orig_query_object = new_query_object
         else
           if !@federation_query_object
-            @federation_query_object = federation_query(@orig_query_object)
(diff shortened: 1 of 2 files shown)


mziwisky:
🦗 guess i was the only one who wanted to do this 🦗

PR #146, opened 2021-12-09
Bug reproduction

Austio:
### Setup

All commands ran at root of project

#### Setup Node
 - install node
 - npm install -g yarn
 - yarn install

#### Setup Ruby 
 - Install Ruby (my example i am running  2.7.2p137, don't think it will matter much if stay below 3)
 - gem install bundler
 - bundle install 
 
#### Run Servers
 - Start subgraph `bundle exec ruby ./example/products.rb`
 - Start gateway `yarn run start-gateway`

### Summary

Expectation: When we pass an invalid variable (null) as an argument (Boolean), i expect the clients to recognize the error and not pass the query to the server and the server to recognize the invalid value and not execute the query.

What is happening: When we pass a `null` value to a variable that should be `Boolean` and is use in an `@include` directive the gateway will pass the query to the subgraph, (ruby in this case).  GraphQL Ruby treats includes as falsy/truthy and then returns the value properly, finally when the Gateway reruns the query result through the schema, the error in the type (null to Boolean) causes the next highest parent that is nullable to be null

### Reproduction

This is the query to send in all requests, only variables change

```graphql
query something($includeThing: Boolean = true) {
  topProducts {
    upc
    name @include(if: $includeThing)
  }
}
```

#### Null includeThing argument nulls out the highest top level nullable field

```
// Variables
{ "includeThing": null }

// Result
{
  "data": {
    "topProducts": null
  }
}
```

#### false only removes the name from the result

```
// Variables
{ "includeThing": false }

// Result
{
  "data": {
    "topProducts": [
      {
        "upc": "1"
      },
      {
        "upc": "2"
      },
      {
        "upc": "3"
      }
    ]
  }
}
```

#### true only removes nothing

```
// Variables
{ "includeThing": true }

// Result
{
  "data": {
    "topProducts": [
      {
        "upc": "1",
        "name": "Table"
      },
      {
        "upc": "2",
        "name": "Couch"
      },
      {
        "upc": "3",
        "name": "Chair"
      }
    ]
  }
}
```

#### Invalid errors cause an error

```
// Variables
{ "includeThing": "something" }

// Result
{
  "error": {
    "errors": [
      {
        "message": "Variable \"$includeThing\" got invalid value \"something\"; Expected type Boolean. Boolean cannot represent a non boolean value: \"something\"",
        "locations": [
          {
            "line": 1,
            "column": 17
          }
        ],
        "extensions": {
          "code": "INTERNAL_SERVER_ERROR",
          "exception": {
```

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -5,3 +5,4 @@ source 'https://rubygems.org'
 gemspec
 
 gem 'graphql', '1.10.10'
+gem 'pry'
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -42,7 +42,7 @@ GEM
     debase-ruby_core_source (0.10.9)
     diff-lcs (1.3)
     erubi (1.9.0)
-    google-protobuf (3.13.0-x86_64-linux)
+    google-protobuf (3.13.0)
     graphql (1.10.10)
     i18n (1.8.5)
       concurrent-ruby (~> 1.0)
@@ -115,6 +115,7 @@ DEPENDENCIES
   appraisal
   debase
   graphql (= 1.10.10)
+  pry
   pry-byebug
   rack
   rake
--- a/example/gateway.js
+++ b/example/gateway.js
@@ -3,18 +3,20 @@ const { ApolloGateway } = require('@apollo/gateway');
 
 const gateway = new ApolloGateway({
   serviceList: [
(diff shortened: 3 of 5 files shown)


Austio:
Sorry, didn't mean to make this PR against the main repo.

PR #150, opened 2022-01-11
feat: Remove erronous? lock to x86_64-linux

### LINKED ISSUE #144: Compatibility with Mac M1's
When trying to do rails:db migrate, I encounter this error message due to apollo-federation gem:

rails aborted!
LoadError: cannot load such file -- google/protobuf_c

google/protobuf is the issue, but not sure how to solve this
https://github.com/protocolbuffers/protobuf/issues/8062

daemonsy:
# What's up? 

Hoping this will resolve #144 , where it seems like folks using M1 might have issue with the platform being forced to use the linux binary. Also updates bundler to the latest/greatest, which keeps turning up when you search for M1 compatibility issues, though not directly related to this one. 

Hopefully the bundler update accounts for the fact there are multiple mac platform types better. 

This platform change was introduced by semantic release bot in https://github.com/Gusto/apollo-federation-ruby/commit/631ebfe4bf4fb2d4097f6322be321b8814755234#diff-89cade48462044ee1b672dc5f4c3ec250fbd29effcd8932096a23c1283c6731fR44 which could be a bundler bug :shrug:

### DIFF
--- a/apollo-federation.gemspec
+++ b/apollo-federation.gemspec
@@ -28,7 +28,7 @@ Gem::Specification.new do |spec|
 
   spec.add_dependency 'graphql', '>= 1.9.8'
 
-  spec.add_runtime_dependency 'google-protobuf', '~> 3.7'
+  spec.add_runtime_dependency 'google-protobuf', '~> 3.13.0'
 
   spec.add_development_dependency 'actionpack'
   spec.add_development_dependency 'appraisal'
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -2,7 +2,7 @@ PATH
   remote: .
   specs:
     apollo-federation (1.1.5)
-      google-protobuf (~> 3.7)
+      google-protobuf (~> 3.13.0)
       graphql (>= 1.9.8)
 
 GEM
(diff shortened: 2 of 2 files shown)


apollo-federation.gemspec
```diff
@@ -28,7 +28,7 @@ Gem::Specification.new do |spec|
 
   spec.add_dependency 'graphql', '>= 1.9.8'
 
-  spec.add_runtime_dependency 'google-protobuf', '~> 3.7'
+  spec.add_runtime_dependency 'google-protobuf', '~> 3.13.0'
```
daemonsy, line 31: Wanted to only keep the minor to `3.13` since this PR is only about removing the platform lock. 

Also guessing typically if we wanted to lock versions, having it locked to patch and having the maintainers manually bump minor/major versions is more intention revealing.

daemonsy:
Kudos to @toreriklinnerud for debugging this with me. More context, bundler 2.20 introduced the specific platform locking and may have contributed to this. For more context: 

https://github.com/rubygems/rubygems/issues/4269#issuecomment-758564690
https://github.com/rubygems/rubygems/pull/4127

geshwho (approved):
LGTM

PR #160, opened 2022-01-26
feat: Get Apollo Federation to work with GraphQL 1.13.x

daemonsy:
This is a PR to gauge the effort to get the library working with GraphQL 1.13. 

## The flow and tasks: 

- Pre upgrade work or blockers 
  - [x]  Remove GraphQL1.9.x support
  - [x] Upstream specs that should pass in all versions
  - [x] ~~Fix semantic release issue that keeps changing google-protobuf to a platform specific version #169~~ This is a monster on its own :( 
- [x] Merge required code changes
- [x] Discuss deprecations and removals (e.g. I think it's wise to remove support for GraphQL 1.9 at this point) 
- [x] Test it in internal app 
- ~~[ ] Release new major version of the gem~~ (there's no breaking changes, in fact no code changes, so we should be good to release a minor version)


lib/apollo-federation/has_directives.rb
```diff
@@ -9,7 +9,7 @@ def add_directive(name:, arguments: nil)
       @federation_directives << { name: name, arguments: arguments }
     end
 
-    def to_graphql
+    def to_graphql(*args, **kwargs)
```
daemonsy, line 12: Self note: Check out the definition change in `#to_graphql` and how this method is used

gemfiles/graphql_1.13.gemfile.lock
```diff
@@ -0,0 +1,126 @@
...
+  pry-byebug
+  rack
+  rake
+  rspec
+  rubocop (~> 0.72.0)
+  rubocop-rspec
+  ruby-debug-ide
+
+BUNDLED WITH
+   2.1.4
```
daemonsy, line 126: We should use bundler 2.3.x

spec/apollo-federation/service_field_spec.rb
```diff
@@ -408,75 +406,64 @@ def self.visible?(context)
       )
     end
 
-    context 'with a filter' do
+    context 'context in schema generation' do
```
daemonsy, line 409: This test broke in 1.13.x, because there's a [new validation](https://github.com/rmosolgo/graphql-ruby/blob/a7daf8fc1fc6e626a8597c4a5e978fddf1fdafb9/lib/graphql/static_validation/rules/query_root_exists.rb#L8) in GraphQL Ruby that warns against having an empty schema, but we were asserting that "Field '_service' doesn't exist on type 'Query'". 

After looking at the [original PR](https://github.com/Gusto/apollo-federation-ruby/pull/50) that introduced this, I think the intent is to make sure that `Schema.federation_sdl` respects context and uses it generation. 

See https://graphql-ruby.org/authorization/visibility.html

I struggled with fixing it trying to get the right indirect error message across different GraphQL versions. Also, I think schema level filters are deprecated? I can't find docs on it anymore on [graphql-ruby](https://github.com/rmosolgo/graphql-ruby). 

So with that, I rewrote the test to basically assert that we're using the context in SDL generation. Further improvements welcome.

daemonsy (commented):
Initial self review

Gemfile.lock
```diff
@@ -42,7 +42,7 @@ GEM
     debase-ruby_core_source (0.10.9)
     diff-lcs (1.3)
     erubi (1.9.0)
-    google-protobuf (3.19.3-x86_64-linux)
+    google-protobuf (3.19.3-x86_64-darwin)
```
daemonsy, line 45: :warning: on why there is the platform lock

geshwho (approved):
LGTM!

daemonsy:
@grxy and @geshwho [Public convo] I'm going to test this on an internal app, if it looks good and y'all are good with the code changes, will :shipit:. 

Looking at the changes, it looks like it should be a minor bump, no breaking changes. Folks using GraphQL 1.13 will find it working correctly as expected.

Issue #161, opened 2022-01-26
Remove support for GraphQL 1.9.x


daemonsy:
- It creates additional complexity in this codebase 
- It's pretty old at this point and has a lot of breaking changes compared to GraphQL 1.13.x

PR #165, opened 2022-01-26
fix: Remove disclaimer and change badge to branch main

### LINKED ISSUE #118: Readme Beta Comment
:wave: Hello, thank you for this fantastic contribution to graphql ruby. 

Is this system actually in "beta" mode as the README state?  I have used this in a relatively complex scenario and it seems to be fully spec compliant.  

If that is the case, would you accept a PR to remove the beta declaration out of the README?

Thanks again!

daemonsy:
I realized I did the previous commit wrongly #161 where we are supposed to have the `BREAKING CHANGE: <reason>` label in the footer of the commit message. 

This should be a docs PR, but noob with semantic release, not sure if it will pick up the change and release the gem. 

- Changes the CircleCI badge to `main` branch 
- Removes beta badge and closes #118 

I'd like to fully remove the disclaimer section if we can switch to using GraphQL Ruby's client directives.


geshwho (approved):
nit: The title can also be updated to remove the "change badge to branch main" part.

Otherwise, LGTM!

PR #166, opened 2022-01-27
fix: Bump protobuf to 3.19 and lock on minor version

daemonsy:
1. In https://github.com/Gusto/apollo-federation-ruby/pull/150, the intention was to remove the platform lock. In that PR, `google-protobuf` was locked to `3.13.x` as per the gemspec and specs. 

2. However in reality, consumers could have been installing `3.19.x` which I think is very common. We should relax the lock to minor version which lets people bump to the latest minor version. 

3. But I've also bumped the requirement to `~> 3.19` because of this [CVE](https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2021-22569) 

This should result in a bump to `2.1.0` and that should be alright because of point 2 and if you're locked to below protobuf `3.19`, the major release should guard that

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -2,7 +2,7 @@ PATH
   remote: .
   specs:
     apollo-federation (2.0.0)
-      google-protobuf (~> 3.13.0)
+      google-protobuf (~> 3.19)
       graphql (>= 1.10.14)
 
 GEM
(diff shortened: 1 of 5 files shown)


geshwho (approved):
LGTM

PR #172, opened 2022-02-02
chore: Upstream spec from GraphQL 1.13 branch

daemonsy:
This upstreams 2 spec files from https://github.com/Gusto/apollo-federation-ruby/pull/160 to main, since they pass across other versions. 

Notes in the code review

### DIFF
--- a/spec/apollo-federation/entities_field_spec.rb
+++ b/spec/apollo-federation/entities_field_spec.rb
@@ -71,10 +71,8 @@ RSpec.describe ApolloFederation::EntitiesField do
 
         it 'sets the Query as the owner to the _entities field' do
           expect(
-            schema.graphql_definition
-              .types['Query']
+            schema.query
               .fields['_entities']
-              .metadata[:type_class]
               .owner.graphql_name,
           ).to eq('Query')
         end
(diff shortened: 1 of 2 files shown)


spec/apollo-federation/entities_field_spec.rb
```diff
@@ -71,10 +71,8 @@
 
         it 'sets the Query as the owner to the _entities field' do
           expect(
-            schema.graphql_definition
-              .types['Query']
+            schema.query
               .fields['_entities']
-              .metadata[:type_class]
```
daemonsy, line 77: `.metadata` is deprecated, this was failing on 1.13.x

spec/apollo-federation/service_field_spec.rb
```diff
@@ -408,75 +406,64 @@ def self.visible?(context)
       )
     end
 
-    context 'with a filter' do
+    context 'with context in schema generation' do
```
daemonsy, line 409: This spec is testing that schema generation (`#federation_sdl`) respects a passed in context. But we tested this indirectly by using a filter (`PermissionWhitelist`). I think using a filter class like this is deprecated. 

With the above, just decided to rewrite the spec to make it more direct.

grxy (approved):

Issue #151, opened 2022-01-20
Graphql 1.13.4+


Fadi25:
Hello Gusto,

I'm facing an issue when using the latest Graphql gem version.

It can be reproduced by changing the graphql version to (1.13.4 or 1.13.5)
rspec ./spec/apollo-federation/entities_field_spec.rb:72 

Thanks

daemonsy:
Hey @Fadi25 looking at our appraisals file, we don't have support for GraphQL 1.1.3.x locked down yet. It's one of the higher priority items on my queue to look into. 

Also see https://github.com/Gusto/apollo-federation-ruby/pull/147, I think that's a good place we can start.

daemonsy:
Done a [spike](https://github.com/Gusto/apollo-federation-ruby/pull/160 ) on GraphQL 1.13.x support. It looks like the only real lift is to remove support from GraphQL 1.9.x, which cleans up the codebase quite a little too.  

https://github.com/Gusto/apollo-federation-ruby/issues/161

daemonsy:
Hi @Fadi25, we shipped "official support" for GraphQL 1.13.x. It's essentially just running specs against GraphQL 1.13.x. 
Initially I thought we will be fixing some deprecations, but they went away with removing support for GraphQL 1.9.x. 

I tried running our current specs against 1.13.4 and 1.13.5, they all passed. 

Could you test if the latest version of the library works for you? If not, could you post more information to help us reproduce the problem?

Fadi25:
Thanks for the update @daemonsy,

Indeed tests are green now, however I tried updating gql to 1.13.5 and Apollo-federation 2.2.0 on my project, unfortunately I'm getting this error, for example after running mutation such as

      mutation {
        createRule(input: { threshold: 123 })
        { rule { id } }
      }

   Failure/Error:
       def to_graphql
         field_defn = super # Returns a GraphQL::Field
         field_defn.metadata[:federation_directives] = @federation_directives
         field_defn
       end
     
     ArgumentError:
       wrong number of arguments (given 1, expected 0)
     # /Users/fadi/.rvm/gems/ruby-3.1.0/gems/apollo-federation-2.2.0/lib/apollo-federation/has_directives.rb:12:in `to_graphql'
     # /Users/fadi/.rvm/gems/ruby-3.1.0/gems/graphql-1.13.5/lib/graphql/schema/object.rb:137:in `block in to_graphql'
[...]

PR #177, opened 2022-03-07
fix: Remove to_graphql and make the interpreter runtime a requirement for older GraphQL versions

daemonsy:
# What's up? 

Trying to deal with the problem caused by `to_graphql` in https://github.com/Gusto/apollo-federation-ruby/issues/151. 

## The cause

GraphQL Ruby is deprecating the `to_graphql` method, slated for removal in 2.0. In 1.13, an argument was introduced `def to_graphql(silence_deprecation_warning: false)`. 

We have a `to_graphql` method in our codebase to add federation directives decorations. It is breaking because we defined it as a method with no arguments. 

Tests didn't break because we don't actually call this method in our specs. We rely on GraphQL ruby to call it in GraphQL versions where the **interpreter runtime is not the default.**

## Complications 

First thought was to add a spec to it, proxy arguments over. I'm used to proxying args `Ruby 2.7.x` style. It failed in our specs because the specs are running on `Ruby 2.6.x`. See this [article](https://www.ruby-lang.org/en/news/2019/12/12/separation-of-positional-and-keyword-arguments-in-ruby-3-0/). 

Given that Ruby 3.x is out, our keyword proxying should do the right thing. We also really didn't want to check for arity. 

## Let's just remove `to_graphql`

Specs fail if we remove the method, but why? The library requires the call to populate metadata on the legacy runtime. 

If we made the interpreter runtime a requirement on older versions of GraphQL where it wasn't the default, all specs pass. 

Given how old these versions are, I think this is the best solution. We're going to remove its use for 2.0 anyway.


example/products.rb
```diff
@@ -58,6 +58,11 @@ def top_products(first:)
 end
 
 class ProductSchema < GraphQL::Schema
+  if Gem::Version.new(GraphQL::VERSION) < Gem::Version.new('1.12.0')
```
daemonsy, line 61: All the other test schemas already have this code block, these two just happened to be missed 🤷‍♂️

daemonsy:
I'd think we can remove support for GraphQL < 1.12 once we support 2.0. But I'd like to get some stats first if I can to confirm that they're not widely used by the community.

grxy (approved):
LGTM

PR #180, opened 2022-03-10
fix: remove InstanceDefinable reference

daemonsy:
# What's up? 
`InstanceDefinable` is used for the old block based `define` syntax. It's been deprecated since `1.10`. Not sure why we're referencing it. 

https://github.com/rmosolgo/graphql-ruby/pull/2680

### DIFF
--- a/gemfiles/graphql_2.0.gemfile.lock
+++ b/gemfiles/graphql_2.0.gemfile.lock
@@ -1,9 +1,9 @@
 PATH
   remote: ..
   specs:
-    apollo-federation (2.2.0)
+    apollo-federation (2.2.1)
       google-protobuf (~> 3.19)
-      graphql (>= 2.0)
+      graphql (>= 1.10.14)
 
 GEM
   remote: https://rubygems.org/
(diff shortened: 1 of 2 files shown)


geshwho (commented):
What's up with the gemfile lock changes?

daemonsy:
In the previously merged PR https://github.com/Gusto/apollo-federation-ruby/pull/176 I think after removing a bump in `Gemfile`, the author didn't bundle again, so the lockfile was still in the previous state. 

I didn't catch that.

daemonsy:
Added an issue to track lockfile drift.

geshwho (approved):

modosc:
hi, any eta on getting this fix released?

geshwho:
> hi, any eta on getting this fix released?

@modosc apologies, it looks like this wasn't automatically released because the commit on  `main` did not follow the format that semantic release was expecting. We're still getting used to this system :sweat_smile:

I'll see if I can manually push a release including this fix today or tomorrow.

geshwho:
@modosc Thanks for your patience. As the bot mentioned, this fix should be available on v2.2.2!

PR #185, opened 2022-03-15
fix: Add a ruby verison file

geshwho:
Pins the ruby version to 2.7.5 for development. In the future we should
consider adding a test matrix of officially supported version of ruby.

I am partially doing this in order to cut a release of the gem 😅


daemonsy (approved):
Oh thank you, this has been so painful!

PR #187, opened 2022-03-23
fix: camelize string fields to match sym behavior

geshwho:
Followup to a previous feature release, which intentionally did not camelize string inputs in order to preserve backwards compatibility. Here we are breaking compatibility to make the behavior consistent. Users can revert the behavior using the `camelize` option if it is necessary.

BREAKING CHANGE: string fields will be camelized by default rather than passed as is.

### DIFF
--- a/lib/apollo-federation/field_set_serializer.rb
+++ b/lib/apollo-federation/field_set_serializer.rb
@@ -6,20 +6,18 @@ module ApolloFederation
   module FieldSetSerializer
     extend self
 
-    def serialize(fields)
+    def serialize(fields, camelize: true)
       case fields
       when Hash
         fields.map do |field, nested_selections|
-          "#{camelize(field)} { #{serialize(nested_selections)} }"
+          "#{camelize(field, camelize)} { #{serialize(nested_selections, camelize: camelize)} }"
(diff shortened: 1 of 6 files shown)


lib/apollo-federation/field_set_serializer.rb
```diff
@@ -6,29 +6,27 @@ module ApolloFederation
...
         raise ArgumentError, "Unexpected field set type: #{fields.class}"
       end
     end
 
     private
 
-    def camelize(field)
-      GraphQL::Schema::Member::BuildType.camelize(field.to_s)
+    def camelize(field, camelize)
+      camelize ? GraphQL::Schema::Member::BuildType.camelize(field.to_s) : field.to_s
```
daemonsy, line 29: I tested the behavior of this helper. 

Camel_Me => camelMe 
camel_me => camelMe
camelMe => camelMe 
camelme => camelme 

I think this is the behavior we want. All it does is to lower camelcase snaked case stuff. If you had a slightly different camelisation scheme, it doesn't touch it. :+1:

daemonsy (approved):
LGTM :+1: 

Reminder on the breaking change prefix

PR #188, opened 2022-03-23
fix: add linux and darwin platforms to lockfile

geshwho:


daemonsy:
If this gets us stable bundling where it's no longer changing to linux on every bundle :+1:

PR #195, opened 2022-04-26
Add support for the @shareable, @inaccessible and @override directives

col:
This is the first step toward supporting the 2.0 spec.


col:
Moved to https://github.com/Gusto/apollo-federation-ruby/pull/196

Issue #153, opened 2022-01-25
Apollo Federation 2.0 support


daemonsy:

col:
Is anyone working on adding support for Apollo Federation 2.0?

If not, any thoughts or discussions on how this would be approached?

col:
Draft PR that adds support for `@shareable` directive. https://github.com/Gusto/apollo-federation-ruby/pull/195

col:
Updated the PR to include `@override` as well. https://github.com/Gusto/apollo-federation-ruby/pull/195

These changes were surprisingly straight forward but I'm not sure if I'm missing something. 

Do you think we need to warn people if they try to use a v2 directive when running a v1 gateway? I'd be happy to support something like this but we don't currently have any indication of which version they are running.

col:
@daemonsy You can probably close this now issue now.

dariuszkuc:
As of v3.5.0 all federation directives except for `@composeDirective` (https://github.com/Gusto/apollo-federation-ruby/issues/228) are supported by this lib. Should this be closed?

aleung-flexport:
@dariuszkuc what about the @override directive?

PR #196, opened 2022-04-27
feat: Support Federation v2

col:
This PR adds support for the @sharable, @inaccessible and @override directives. I believe this is all that is required to support Apollo Federation v2.0.

Issue: https://github.com/Gusto/apollo-federation-ruby/issues/153

Original PR https://github.com/Gusto/apollo-federation-ruby/pull/195 was closed due to branch rename.


rmosolgo:
:+1: Seems like it should do the job! The only new feature I read about but didn't see in the tests was `@key(..., resolvable: false)`, but it looks like the existing `add_directive(...)` method should cover it.

daemonsy:
Hey @col thanks for submitting this. :clap:.  I personally haven't caught up with the changes in Federation V2. We should be able to put some focus on this in mid month. 

Meanwhile, @lennyburdette if you're around, be nice to get some thoughts from you on this PR.

col:
Good suggestion @lennyburdette! I was never very happy with my initial approach of searching for usages of the v2 directives. 

I've opted for the version to be a keyword arg so that we can add other options in the future if required. 
Example: 
```
class MySchema
  federation version: 2
end
```

lib/apollo-federation/schema.rb
```diff
@@ -12,9 +12,27 @@ def self.included(klass)
     end
 
     module CommonMethods
+      FEDERATION_2_PREFIX = <<~SCHEMA
+        extend schema
+          @link(url: "https://specs.apollo.dev/federation/v2.0",
+          import: ["@key", "@extends", "@external", "@requires", "@provides", "@shareable", "@inaccessible", "@override"])
```
lyricsboy, line 18: This looks like it's importing *all* of the directives unconditionally. Is that acceptable, or should it be more dynamic to only import the directives that are actually used in the schema?

[Apollo docs](https://www.apollographql.com/docs/federation/federated-types/federated-directives/#importing-directives) say "Modify the import array to include whichever federated directives your subgraph schema uses."

lyricsboy: Is `@extends` a valid directive in this context? I don't see it in the docs.

lyricsboy (commented):
Looking forward to seeing this land!

lib/apollo-federation/schema.rb
lennyburdette: One subtle feature of the `@link` spec is that you don't _have_ to import these directives. If you don't supply the `import:` argument, the directives are still available in a namespaced form:

```graphql 
extend schema @link(url: "https://specs.apollo.dev/federation/v2.0")

type Foo @federation__key(fields: "bar") @federation__shareable {
  bar: ID!
  baz: String @federation__external
  quux: String @federation__requires(fields: "bax")
  blah: String @federation__inaccesible
}
```

This might be the "right" approach for a code-first library. The `key` ruby method could output `@federation__key`.

lennyburdette: And yes, I believe the `@extends` directive was dropped entirely in Fed 2.

col: I think `@extends` is still optionally supported. We either need to support `@extends` or have proper support for type extensions (eg.  `extend type`).
See: https://www.apollographql.com/docs/federation/federation-spec/#federation-schema-specification

col: > This looks like it's importing _all_ of the directives unconditionally. Is that acceptable, or should it be more dynamic to only import the directives that are actually used in the schema?
> 
> [Apollo docs](https://www.apollographql.com/docs/federation/federated-types/federated-directives/#importing-directives) say "Modify the import array to include whichever federated directives your subgraph schema uses."

I did consider trying to make this dynamic but it was complicated to implement. In the end, I couldn't think of a use case where this would actually cause a problem so I figured it was ok to include all the directives. @lyricsboy Do you have an example of how this could be a problem?

@lennyburdette - Not a big fan of using the `@federation__external` approach as I've never seen this mentioned in any of the docs. I fear it would cause some confusion when the schema we're generating doesn't match what people see in the docs. If we were to take this approach, should it apply to v1 and v2 or only v2?

spec/apollo-federation/service_field_spec.rb
```diff
@@ -148,6 +148,135 @@ def execute_sdl(schema)
       )
     end
 
+    it 'returns valid SDL for shareable types' do
```
daemonsy, line 151: Given that the spec cases are testing the same things working in Federation 2.0, I suggest that we make a separate `v2` folder for all the v2 test cases? The specs in this case can be named `spec/apollo-federation/v2/service_field_spec.rb`. 

Makes it easier to find them. When the day comes when we don't support 1.0 anymore, it makes clean up easier too.

README.md
```diff
@@ -61,6 +61,15 @@ class MySchema < GraphQL::Schema
 end
 ```
 
+**Optional:** To opt in to Federation v2, specify the version in your schema:
+
+```ruby
+class MySchema < GraphQL::Schema
+  include ApolloFederation::Schema
+  federation version: 2.0
```
daemonsy, line 69: Might this be the place to support `import`?

lib/apollo-federation/schema.rb
daemonsy: The way this is done currently, @lennyburdette , does it pass the `@link` test in https://www.apollographql.com/docs/federation/other-servers/?

daemonsy:
Hi @col @lyricsboy @lennyburdette thanks for all the work and comments on the PR so far.  

@geshwho and I paired over reviewing the 2.0 spec and the PR. Overall the PR looks great! Jotting down a few notes:

### Code considerations
- [ ] In this PR, do we want to improve `@link` support so directives can be renamed?
- [ ] Let's work separately on updating https://github.com/apollographql/apollo-federation-subgraph-compatibility to do V2 tests 

### Ops 

- When reviews are all good, we will ship this a major version bump (technically it's non breaking) 
- Help internal Gusto apps adopt the latest gem version whilst still using V1. This helps to ensure that we didn't break anything in V1 backwards compatibility. FYI @grxy we can discuss separately switching to V2. 
- Mark on the README that beta 2.0 support is available, encourage folks to try it out. 
- Remove beta when we're confident. Ideally, by that time, Gusto is already on V2.

daemonsy (commented):
Some comments

lib/apollo-federation/schema.rb
lennyburdette: Yeah this should be sufficient. The test is rudimentary right now: it just searches the `_Service.sdl` for the presence of "@link": https://github.com/apollographql/apollo-federation-subgraph-compatibility/blob/main/src/tests/link.test.ts

README.md
lennyburdette: `version:` should probably be a string. we don't foresee a need for `2.0.1` but it's possible

lennyburdette:
FYI I spoke with the maintainer of graphql-kotlin (another code-first graphql library) and I think we're settled on emitting only namespaced directives (`@federation__key`) and _not_ providing an API to specify imports or renames. I'm pretty sure this will be the path forward for most code-first libraries.

daemonsy:
Thanks @lennyburdette !

daemonsy:
Hi @col are you still actively working on this PR? No worries if you're not able to, we can merge the existing work and do some of the action items separately.

col:
> Hi @col are you still actively working on this PR? No worries if you're not able to, we can merge the existing work and do some of the action items separately.

I'm happy to incorporate some of the feedback and update the PR over the next day or two.

col:
@daemonsy (CC: @lennyburdette @lyricsboy) I've updated the PR with your feedback. Please take another look.

Changes:
- Changed the federation version to support semantic versioning
- Remove the explicit import of the federation directives
- Output the directives using the `federation__` namespace prefix when using >= 2.0
- Split the service field specs into v1 and v2

daemonsy:
Hey @col it's looking good to me. I am going to run some sanity check on the PR using our codebase. 

@lennyburdette how does it look to you?

lennyburdette:
looks great!

spec/apollo-federation/schema_spec.rb
```diff
@@ -0,0 +1,73 @@
...
+    it 'returns false when version is less than 2.0' do
+      schema = Class.new(GraphQL::Schema) do
+        include ApolloFederation::Schema
+        federation version: '1.5'
+      end
+
+      expect(schema.federation_2?).to be(false)
+    end
+
+    it 'returns true when the version an integer equal to 2' do
```
sethc2, line 46:
```suggestion
    it 'returns true when the version is an integer equal to 2' do
```

sethc2 (changes requested):
Looks good to me. Just a couple grammar nits.

Thanks for putting this together!

pmrotule:
@daemonsy Any update? Can't wait to use this feature 😅 👏

PR #202, opened 2022-08-09
feat: allow custom namespaces for linked directives

geshwho:


lib/apollo-federation/federated_document_from_schema_definition.rb
```diff
@@ -62,13 +62,21 @@ def merge_directives(node, type)
...
-          name: schema.federation_2? ? "federation__#{directive[:name]}" : directive[:name],
+          name: directive_name(directive),
           arguments: build_arguments_node(directive[:arguments]),
         )
       end
       node
     end
 
+    def directive_name(directive)
+      if schema.federation_2?
```
daemonsy, line 73: The PR looks good to me. Is there a case where the user might not want any namespace, aka they just want it to be `directive[:name]`?

geshwho: Yes, per the spec this is done with the import argument:

> The import argument enables you to import external definitions into your namespace, so they are not prefixed. 

This is a separate thing we need to implement.

daemonsy (approved):
LGTM!

Issue #205, opened 2022-08-16
Using a dataloader inside `resolve_reference` does not work


col:
It's important that when a subgraph service receives an `_entities` query from the gateway that it can resolve all of the references efficiently. You would expect that using a data loader inside the `resolve_reference` method for each entity would achieve this but unfortunately it does not work as expected. Currently the data loader is being invoked for each individual reference rather than as a single batch.

I've created a small example application here: [https://github.com/col/apollo-federation-example](https://github.com/col/apollo-federation-example)

Expected Output: 
```
Product.resolve_reference reference: {:__typename=>"Product", :id=>"1"}
Product.resolve_reference reference: {:__typename=>"Product", :id=>"2"}
ProductsByUpc.fetch ["1", "2"]
```

Actual Output:
```
Product.resolve_reference reference: {:__typename=>"Product", :id=>"1"}
ProductsByUpc.fetch ["1"]
Product.resolve_reference reference: {:__typename=>"Product", :id=>"2"}
ProductsByUpc.fetch ["2"]
```

As scale this becomes a significant problem as we can receive hundreds of references in a single `_entities` query.

geshwho:
I'm not quite as well versed on this issue as @daemonsy, but I'll dig into it this week (@daemonsy is currently OOO)

@col Am I correct that this is the same issue (or very similar to) as #149?

col:
@geshwho Yes you are correct, this is the same issue as #149

daemonsy:
Hey @col, should have documented the other issue better, a little hazy now. I believe I didn't actually find a direct solution to the problem, it just wasn't a problem anymore on the project that's using it. 

Had a window to take a look but couldn't get the example to run. I believe it has something to do with how we map `representations` in `#_entities`

rmosolgo:
I think this basically boils down the the difference between GraphQL-Batch and Dataloader: 

- With GraphQL-Batch, user code would return a `Promise` to GraphQL-Ruby (or an Array of Promises), then GraphQL-Ruby would resolve them at some point 
- With Dataloader, `.load(...)` calls actually _pause_ execution until the data is available, then it resumes. 

[...]

PR #206, opened 2022-08-21
feat: introduce optional `resolve_references` method

col:
This addresses the issues with using data loader in the existing `resolve_reference` method. (See: #205) 
By grouping all the references of the same type into a single call to `resolve_references` the records can more easily be loaded as a batch. This can be done using a data loader (`.load_all`) or just a direct DB query as all the references are provided in a single method call. 

The existing `resolve_reference` method is still supported but `resolve_references` will be the preferred approach.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,3 +9,5 @@ node_modules/
 .circleci/processed-config.yml
 
 .DS_Store
+.idea
+vendor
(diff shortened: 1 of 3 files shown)


geshwho:
@col Thank you for the PR! This is looking pretty good to me.

geshwho (approved):

PR #218, opened 2023-01-31
feat: adds support for the `@interfaceObject` directive

moonflare:
### Context
The Apollo team are working on a new release for the Federation spec. This release will bump the spec to `v2.3` and one of the main features is the support for referencing Entities that are interfaces. You can read more about this [here](https://github.com/apollographql/federation/issues/2277), and check the initial implementation [here](https://github.com/apollographql/federation/pull/2279). 

### Description
In this PR I'm addressing two minor things:
* Adds support for the `@interfaceObject` for the `Object` class
* Adds support to link the latest version of `v2.3`
* Fixes the custom namespace not being added to the SDL. (eg: `@link(url: "https://specs.apollo.dev/federation/v2.3", as: "fed2")`)

### Considerations
* We need to add a validation check in `v2.3` is that we can't apply `@key` directive on interfaces for versions lower than `v2.3`. Also, if we do (in `v2.3+`), we need to make sure all the implementing types also have the `@key` directive. (According to [this](https://github.com/apollographql/federation/pull/2279/files#diff-b1fed99113813ea67905b0212fe42e1255e04c152df516f0acf92c7466f721deR48))

_Regarding the consideration above, let me know if you think it makes sense to add it part of this PR or it can be added after we get this one in. Currently, we need the existing changes to support the new features in `v2.3`, but I'm willing to add the extra validations so we have a consistent implementation of the `v2.3` spec support._

### DIFF
--- a/lib/apollo-federation/schema.rb
+++ b/lib/apollo-federation/schema.rb
@@ -12,15 +12,11 @@ module ApolloFederation
     end
 
     module CommonMethods
-      FEDERATION_2_PREFIX = <<~SCHEMA
-        extend schema
-          @link(url: "https://specs.apollo.dev/federation/v2.0")
-
-      SCHEMA
+      DEFAULT_LINK_NAMESPACE = 'federation'.freeze
 
       def federation(version: '1.0', link: {})
         @federation_version = version
-        @link = { as: 'federation' }.merge(link)
+        @link = { as: DEFAULT_LINK_NAMESPACE }.merge(link)
       end
 
       def federation_version
@@ -35,7 +31,7 @@ module ApolloFederation
         document_from_schema = FederatedDocumentFromSchemaDefinition.new(self, context: context)
 
         output = GraphQL::Language::Printer.new.print(document_from_schema.document)
-        output.prepend(FEDERATION_2_PREFIX) if federation_2?
+        output.prepend(federation_2_prefix) if federation_2?
         output
       end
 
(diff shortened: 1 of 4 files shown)


lib/apollo-federation/schema.rb
```diff
@@ -12,30 +12,27 @@ def self.included(klass)
     end
 
     module CommonMethods
-      FEDERATION_2_PREFIX = <<~SCHEMA
-        extend schema
-          @link(url: "https://specs.apollo.dev/federation/v2.0")
-
-      SCHEMA
+      DEFAULT_LINK_NAMESPACE = 'federation'.freeze
+      SUPPORTED_FED2_VERSIONS = ['2.0', '2.1', '2.2', '2.3'].freeze
```
dariuszkuc, line 16: for a code first* library IMHO it does not make sense to support multiple fed 2 versions -> if you use the library with the v2, you probably should just default to the latest

*in SDL first where we load the provided file it might be needed

dariuszkuc:
@grxy can you help out with the review of this PR?

grxy (approved):
This works great! If you could follow up with a PR to add the `@key` support you mentioned, that would be much appreciated. Also, for future PRs, it looks like you will need to allow Gusto CircleCI orbs so that you can run CI. See [this build page](https://app.circleci.com/pipelines/github/moonflare/apollo-federation-ruby) for the failure that prevented it from running.

PR #220, opened 2023-02-21
fix: import `@inaccessible` directive in Fed v2.3

moonflare:
### Description

Currently, due to different namespaces across multiple subgraphs, we are getting the following error:

```GraphQL
LINK_IMPORT_NAME_MISMATCH: The federation "@inaccessible" directive is imported with mismatched name between subgraphs: it is imported as "@federation__inaccessible" in subgraphs "A" and "B" but "@inaccessible" in subgraphs "C", "D"...
```

According to [prasek](https://github.com/prasek)'s comment [here](https://github.com/apollographql/federation/issues/1976#issuecomment-1184522540), we should always import the `@inaccessible` directive to avoid multiple namespaces. This PR address this.

### DIFF
--- a/lib/apollo-federation/federated_document_from_schema_definition.rb
+++ b/lib/apollo-federation/federated_document_from_schema_definition.rb
@@ -14,6 +14,7 @@ module ApolloFederation
       '_entities',
       '_service',
     ].freeze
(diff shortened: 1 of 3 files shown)


grxy (approved):
Thanks!

PR #222, opened 2023-02-23
Respect Federation version in @link directive

lyricsboy:
Problem
---

The federation version, though specifiable in the schema configuration, was not reflected in the `@link` directive URL when output via the `federation_sdl` method.

Solution
---

Use the federation_version as specified when building the URL.

In this implementation, I relied on some otherwise-unused code in the `federation_2_prefix` method.

Note: it's currently in conflict with `main`, but I wanted to illustrate that the tests are passing as written here. If we proceed with this, we can fix it up to incorporate the changes from main related to the inaccessible directive.


dzmitry-miadzvedzeu:
Hello Everyone. 

I want to clarify the status of this PR, if this change is considered a "breaking change", and hence, not possible to merge right now (reference to #267)?

Thank you in advance for your time and assistance.

Kaytal:
I know it's been a while since this PR was originally opened. However, I second this PR. It shouldn't be considered a breaking change any more than defining your `federation version: 2.1` and getting a `@link` spec that matches federation 2.3. Additionally, locking the specification to 2.3 prevents many additional improvements and directives on later versions of Apollo Federation.

PR #231, opened 2023-03-24
chore: Bump google-protobuf dependency

such:
google-protobuf 3.21 does not support ruby 3.2.0

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -2,7 +2,7 @@ PATH
   remote: .
   specs:
     apollo-federation (3.6.0)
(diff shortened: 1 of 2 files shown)


grxy (dismissed):
~Thanks for the PR! Can you add a test that demonstrates the failing functionality in Ruby 3.2? This PR inspired me to create [another PR](https://github.com/Gusto/apollo-federation-ruby/pull/232) to ensure we are testing against multiple Ruby versions.~ Nvm. Was able to repro myself. Going to incorporate in a separate PR and tag you. Thanks again!

Issue #217, opened 2023-01-17
Add `resolvable` argument to `key` directive


BPMathews:
Apollo documentation describes how to [reference an entity without contributing fields](https://www.apollographql.com/docs/federation/entities/#referencing-an-entity-without-contributing-fields). The `resolvable` argument on the `key` field indicates to Apollo that the subgraph is only referencing the entity and has no resolver. 

I looked at the key implementation [here](https://github.com/Gusto/apollo-federation-ruby/blob/fc541eec17768db614efc62d3d0217eace9bf354/lib/apollo-federation/object.rb#L27) and didn't see a way set this argument in my subgraph. 

Could this be added in order to fulfill this aspect of federation? 

**Schema Example**
```
type Product @key(fields: "id", resolvable: false) {
  id: ID!
}
```

grxy:
@BPMathews Thanks for pointing this out! Feel free to open a PR if you're up to it! CC: @sethc2

PR #242, opened 2023-04-18
feat: add resolvable argument to key annotation

Goos:
This adds support for federation 2.0's new [resolvable field](https://www.apollographql.com/docs/federation/federation-2/new-in-federation-2/#changes-to-key).

This was referred to [here](https://github.com/Gusto/apollo-federation-ruby/issues/217), but I never saw a fix for it, so here it is.

### DIFF
--- a/README.md
+++ b/README.md
@@ -178,6 +178,14 @@ class User < BaseObject
 end
 ```
 
+As well as non-resolvable keys:
+
+```ruby
+class User < BaseObject
+  key fields: :id, resolvable: false
+end
+```
+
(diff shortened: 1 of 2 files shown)


jturkel:
We hit this problem too and would love a fix. @Goos - want me to add a test for you?

grxy:
@jturkel If we don't hear back, we may be able to incorporate these changes + your tests into a separate PR and tag @Goos as a co-author.

jturkel:
I branched off this PR and created #249 with the tests.

grxy:
This was incorporated into #249.

Issue #197, opened 2022-05-19
Enabling federated tracing breaks multiplex


ivanbrennan:
In a Rails app with a controller that supports [query-batching via multiplex](https://graphql-ruby.org/queries/multiplex.html#apollo-query-batching), enabling federated tracing ([using the documented steps](https://github.com/Gusto/apollo-federation-ruby#tracing)) leads to a runtime error when processing batched queries.

#### Repro

Controller action:
```ruby
  def execute
    context = {
      tracing_enabled: ApolloFederation::Tracing.should_add_traces(request.headers),
    }

    multi = params[:_json]
    result =
      if multi
        MySchema.multiplex(
          multi.map do |param|
            {
              query: param[:query],
              operation_name: param[:operationName],
              variables: prepare_variables(param[:variables]),
              context: context
            }
          end
        )
      else
        MySchema.execute(
          params[:query],
          operation_name: params[:operationName],
          variables: prepare_variables(params[:variables]),
          context: context
        )
      end

    render json: result
  end
```

The following requests succeed:
```sh
# single query
$ curl localhost:3000/graphql -H 'content-type: application/json' -d '{"query":"query{__schema{__typename}}"}'
{"data":{"__schema":{"__typename":"__Schema"}}}

# batch of 1
$ curl localhost:3000/graphql -H 'content-type: application/json' -d '[{"query":"query{__schema{__typename}}"}]'
[{"data":{"__schema":{"__typename":"__Schema"}}}]
```

The following request fails:
```sh
# batch of 2
$ curl localhost:3000/graphql -H 'content-type: application/json' -d '[{"query":"query{__schema{__typename}}"},{"query":"query{__schema{__typename}}"}]'
```
```
Completed 500 Internal Server Error in 1831ms (Allocations: 7241)

NoMethodError (undefined method `context' for nil:NilClass):

apollo-federation (3.0.0) lib/apollo-federation/tracing/tracer.rb:87:in `execute_query_lazy'
apollo-federation (3.0.0) lib/apollo-federation/tracing/tracer.rb:48:in `trace'
graphql (2.0.5) lib/graphql/tracing.rb:82:in `call_tracers'
graphql (2.0.5) lib/graphql/tracing.rb:66:in `trace'
graphql (2.0.5) lib/graphql/execution/interpreter.rb:71:in `sync_lazies'
graphql (2.0.5) lib/graphql/execution/interpreter.rb:32:in `finish_multiplex'
[...]

PR #243, opened 2023-04-21
fix(tracer): add support for multiplex

ashleyyy:
Fixes https://github.com/Gusto/apollo-federation-ruby/issues/197

We found that once we implemented tracing, multiplex queries would fail with only a `NoMethodError (undefined method 'context' for nil:NilClass)`

In debugging I found that multiplex queries end up hitting `Tracer.execute_query_lazy` which was expecting a single query at `data[:query]`, which was `nil` when `data[:multiplex].queries` included more than one query. 

This PR adds support for either a single query or several queries as in a multiplex context.


grxy (approved):
Thanks for the fix!

PR #244, opened 2023-05-14
fix: call `resolve_references` once per `__typename`

zachahn:
This PR is a follow up to #239.

The previous implementation called `resolve_references` once per contiguous `__typename` to ensure the correct ordering of the response. This could have resulted in wildly varying runtime efficiencies. A request calling for _N_ unique types across _M_ entities could have as little as _N_ queries or as many as _M_ queries, depending on if the entities happened to be in large groups of the same type or very interleaved.

By keeping track of the requested representations and their original indices, we can rebuild the response in the correct order even if things get shuffled during computation.

### DIFF
--- a/lib/apollo-federation/entities_field.rb
+++ b/lib/apollo-federation/entities_field.rb
@@ -27,9 +27,17 @@ module ApolloFederation
     end
 
     def _entities(representations:)
-      chunked_references = representations.chunk { |r| r[:__typename] }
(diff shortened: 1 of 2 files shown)


zachahn:
Hi @grxy, just pinging you since you reviewed my last PR that touched this code. Had some time this past weekend to work on the performance changes. If you get a chance, let me know what you think!

zachahn:
Thank you!

PR #245, opened 2023-05-19
fix: visibility checks should handle duplicate types and fields

jturkel:
This PR fixes two visibility scenarios that graphql-ruby supports but apollo-federation-ruby's does not:

1. Multiple object types with the same name as long as only one is visible at runtime
2. Multiple fields with the same name as long as only one is visible at runtime

The first scenario was broken because visibility checking in [GraphQL::Schema.types](https://github.com/rmosolgo/graphql-ruby/blob/3a548f8cb540f16f09506bde9fc0778d3eba276d/lib/graphql/schema.rb#L286) was applied when generating the `_Entity` union but we don't have a properly setup context at this point, resulting in either a `GraphQL::Schema::DuplicateNamesError` or a `NoMethodError: undefined method 'introspection?' for nil:NilClass` depending on how an application has setup their visibility checks for types with the same name. The fix is to avoid any visibility checks when generating the `_Entity` union and leave it up to the graphql-ruby runtime to apply visibility checks during query execution when we have a proper context.

The second scenario was broken because the federation SDL generation was looking at all of a type's fields rather than just its visible fields resulting in `GraphQL::Schema::DuplicateNamesError`. The fix is to use warden to only look at the visible fields.

### DIFF
--- a/spec/apollo-federation/service_field_v2_spec.rb
+++ b/spec/apollo-federation/service_field_v2_spec.rb
@@ -2043,6 +2043,76 @@ RSpec.describe ApolloFederation::ServiceField do
         )
       end
     end
+
+    if Gem::Version.new(GraphQL::VERSION) >= Gem::Version.new('1.13.0')
+      context 'with visibility checks on types and fields with duplicate names' do
+        let(:schema) do
+          regular_product = Class.new(base_object) do
+            graphql_name 'Product'
+            key fields: :upc
+
+            field :upc, String, null: false
+            field :regular_field, String, null: true
+
+            def self.visible?(context)
+              context[:graph_type] == :regular
+            end
+          end
+
+          admin_product = Class.new(base_object) do
+            graphql_name 'Product'
+            key fields: :upc
+
+            field :upc, String, null: false
+            field :admin_field, String, null: true
+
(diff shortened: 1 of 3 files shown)


jturkel:
🤔 Looks like there's a problem with my unit tests in graphql-ruby <= 1.12. Investigating...

jturkel:
It looks like graphql-ruby added support for these scenarios in [1.13.0](https://github.com/rmosolgo/graphql-ruby/blob/master/CHANGELOG.md#1130-24-november-2021) so I updated the unit test to only target 1.13.0+.

spec/apollo-federation/service_field_v2_spec.rb
```diff
@@ -2043,6 +2043,76 @@ def hello
...
+              end
+            end
+          end
+
+          Class.new(base_schema) do
+            query query_obj
+          end
+        end
+
+        it 'applies visibility checks during SDL generation' do
```
grxy, line 2097: Can you add a test for `context[:graph_type] == :admin` as well?

grxy (changes requested):
Requesting one additional test.

grxy (approved):

PR #248, opened 2023-05-24
feat: Optionally underscore _Any keys

jturkel:
This PR adds optional support to underscore the keys in the `_Any` scalar for consistency with the `fields` argument of the `ApolloFederation::Object.key` method and graphql-ruby's general conversion from camelcase to snakecase. I disabled this by default for backwards compatibility but it should probably be enabled by default in the next major release.

### DIFF
--- a/README.md
+++ b/README.md
@@ -325,7 +325,8 @@ class User < BaseObject
   end
 end
 ```
-By default, reference hash keys are camelcase but they can be underscored by setting `underscore_reference_keys` on your entity class:
+
+To maintain backwards compatibility, by default, reference hash keys are camelcase. They can be underscored by setting `underscore_reference_keys` on your entity class. In order to maintain consistency with GraphQL Ruby, we may change the keys to be underscored by default in a future major release.
 
 ```ruby
 class User < BaseObject
--- a/README.md
+++ b/README.md
@@ -178,6 +178,14 @@ class User < BaseObject
 end
 ```
 
+As well as non-resolvable keys:
+
+```ruby
+class User < BaseObject
+  key fields: :id, resolvable: false
+end
+```
+
 See [field set syntax](#field-set-syntax) for more details on the format of the `fields` option.
 
 ### The `@external` directive
--- a/README.md
+++ b/README.md
@@ -317,12 +317,37 @@ Define a `resolve_reference` class method on your object. The method will be pas
 
 ```ruby
 class User < BaseObject
+  key fields: :user_id
+  field :user_id, ID, null: false
+  
   def self.resolve_reference(reference, context)
-    USERS.find { |user| user[:id] == reference[:id] }
+    USERS.find { |user| user[:userId] == reference[:userId] }
   end
 end
 ```
-By default, reference hash keys are camelcase but they can be underscored by setting `ApolloFederation::Any.underscore_keys = true`.
+By default, reference hash keys are camelcase but they can be underscored by setting `underscore_reference_keys` on your entity class:
+
+```ruby
+class User < BaseObject
+  key fields: :user_id
+  field :user_id, ID, null: false
+  underscore_reference_keys true
+  
+  def self.resolve_reference(reference, context)
+    USERS.find { |user| user[:user_id] == reference[:user_id] }
+  end
+end
+```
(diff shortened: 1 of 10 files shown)


grxy:
@jturkel Thanks for the PR. This could lead to a big migration for large projects. Is there any way we could make this adoptable on a per-entity basis?

jturkel:
@grxy - That's a fair concern. What about an `underscore_reference_keys` DSL method on `ApolloFederation::Object` that lets individual entities opt into this behavior e.g.

```ruby
class User < BaseObject
  underscore_reference_keys true

  def self.resolve_reference(reference, context)
    USERS.find { |user| user[:my_id] == reference[:my_id] }
  end
end
```

This would also allow applications to opt into this behavior for all entities via a `BaseObject` superclass:

```ruby
class BaseObject < GraphQL::Schema::Object
  include ApolloFederation::Object

  field_class BaseField
  underscore_reference_keys true
end
```

grxy:
@jturkel I think that approach makes a lot more sense since it allows both incremental and all-at-once adoption. We can talk about it in the future, but we can probably make that behavior the default in a future major release, for consistency with GraphQL Ruby.

jturkel:
@grxy - I switched to using a per-entity setting. Let me know what you think.

lib/apollo-federation/entities_field.rb
```diff
@@ -49,6 +49,14 @@ def _entities(representations:)
         # TODO: What if the type is an interface?
         type_class = class_of_type(type)
 
+        if type_class.underscore_reference_keys
+          representations.each do |representation|
+            representation.transform_keys! do |key|
+              GraphQL::Schema::Member::BuildType.underscore(key.to_s).to_sym
+            end
+          end
```
grxy, line 57: We might want to transform a copy instead of the original representations. Could cause weird issues down the road if we do this destructively.

README.md
```diff
@@ -317,11 +317,37 @@ Define a `resolve_reference` class method on your object. The method will be pas
...
+  key fields: :user_id
+  field :user_id, ID, null: false
+  
   def self.resolve_reference(reference, context)
-    USERS.find { |user| user[:id] == reference[:id] }
+    USERS.find { |user| user[:userId] == reference[:userId] }
   end
 end
 ```
+By default, reference hash keys are camelcase but they can be underscored by setting `underscore_reference_keys` on your entity class:
```
grxy, line 328: I would change this to say something like:

```
To maintain backwards compatibility, by default, reference hash keys are camelcase. They can be underscored by setting `underscore_reference_keys` on your entity class. In order to maintain consistency with GraphQL Ruby, we may change the keys to be underscored by default in a future major release.
```

spec/apollo-federation/entities_field_spec.rb
```diff
@@ -475,6 +475,59 @@ def load_entity
                   end
                 end
               end
+
+              context 'when reference keys have multiple words' do
```
grxy, line 479: Can we add one more test covering the inheritance of `underscore_reference_keys`, such that it it enabled in the base object and validates that it still works in a child object class?

grxy (changes requested):
This seems almost there. Have a suggestion for documentation, a small implementation detail change, and an additional test request.

lib/apollo-federation/entities_field.rb
```diff
@@ -49,6 +49,14 @@ def _entities(representations:)
         # TODO: What if the type is an interface?
         type_class = class_of_type(type)
 
+        if type_class.underscore_reference_keys
+          references.map! do |reference|
```
grxy, line 53: Does this not still mutate? Could we do something like `references = references.map` here to avoid changing anything about the original references array or its members?

jturkel: This is mutating the local `references` array allocated a few lines up on [this](https://github.com/Gusto/apollo-federation-ruby/pull/248/files#diff-353631238a3878fef788c3ab953b5d493aabdb176b7a23b52deda2b305955e6fL38) line. The original `representations` method argument and its members aren’t modified with these latest changes.

grxy: Oops. Missed that. Thanks for pointing it out.

grxy (approved):

Issue #221, opened 2023-02-23
v4.3.x updates federation directive version to 2.3, fails schema checks with Apollo Studio


lyricsboy:
Howdy! We updated this gem to 4.3.1 in one of our subgraphs, and its schema verification is failing in Apollo Studio. Here are some results:

Generated Schema snippet:

```
extend schema
  @link(url: "https://specs.apollo.dev/federation/v2.3", import: ["@inaccessible"])

type Something @federation__extends @federation__key(fields: "id") {
  id: ID! @federation__external
  ...
}
```

Apollo Studio Check output:

```
UNKNOWN_FEDERATION_LINK_VERSION
[x-service] Invalid version v2.3 for the federation feature in @link direction on schema
INVALID_GRAPHQL
[x-service] Unknown directive "@federation__external".
INVALID_GRAPHQL
[x-service] Unknown directive "@federation__extends".
INVALID_GRAPHQL
[x-service] Unknown directive "@federation__key".
```

Our supergraph is configured in Apollo Studio to use Federation version 2.0.

Our subgraphs also specify the Federation version using this gem. It seems like that version specification is being ignored when generating the `@link` directive.

I'm looking at this code change 👀  https://github.com/Gusto/apollo-federation-ruby/commit/c7b987de1d2b32a4a77ceb09718373ffa5a60abb#diff-48d1e38dfab06335bc60255c48ef179d40143d14e29cfb52673b258346799900R64

Shouldn't that be honoring the `federation_version` value?

grxy:
CC: @moonflare @dariuszkuc

lyricsboy:
I opened https://github.com/Gusto/apollo-federation-ruby/pull/222 as a proposed fix for this.

grxy:
I think you may be able to fix this by updating your selected version of composition in Apollo Studio. Have you tried that yet? I think your PR could work as well, but would be great if there is a path to unblocking without requiring a code change. Let me know!

lyricsboy:
I can try that. We can unblock ourselves by pinning to a prior release of this gem, as well. 

Generally, though, is it correct that users of this gem should not be required to use a particular version of Federation other than what they specify in their own setup code? Like, we shouldn't be expected to use 2.3 for composition just because it's hard-coded in this gem, right?

dariuszkuc:
In regards to the failure -> new version of a federation spec (v2.3) requires usage of compatible composition version (in studio) and a router/gateway that supports it. General upgrade path should be

[...]

PR #253, opened 2023-06-16
Add support for @composeDirective

utay:
Fixes https://github.com/Gusto/apollo-federation-ruby/issues/228

This PR adds support for `@composeDirective` as well as having multiple `@link` on the schema which is a prerequisite of the former. There's a breaking change `link`  -> `default_link_namespace` in the `federation` function because I felt like it makes more sense like this. Happy to hear your thoughts!

### DIFF
--- a/lib/apollo-federation/schema.rb
+++ b/lib/apollo-federation/schema.rb
@@ -8,6 +8,7 @@ require 'apollo-federation/federated_document_from_schema_definition.rb'
 module ApolloFederation
   module Schema
     IMPORTED_DIRECTIVES = ['inaccessible', 'tag'].freeze
+    IMPORTED_DIRECTIVES_V2_1 = ['composeDirective'].freeze
 
     def self.included(klass)
       klass.extend(CommonMethods)
@@ -16,9 +17,16 @@ module ApolloFederation
     module CommonMethods
       DEFAULT_LINK_NAMESPACE = 'federation'
 
-      def federation(version: '1.0', link: {})
+      def federation(version: '1.0', default_link_namespace: nil, links: [], compose_directives: [])
         @federation_version = version
-        @link = { as: DEFAULT_LINK_NAMESPACE }.merge(link)
+        @default_link_namespace = default_link_namespace
+        @links = links
+
+        if !federation_2_1? && compose_directives.any?
+          raise ArgumentError, 'composeDirective is available in Federation 2.1 and later'
+        end
+
+        @compose_directives = compose_directives
       end
 
       def federation_version
@@ -29,6 +37,10 @@ module ApolloFederation
         Gem::Version.new(federation_version.to_s) >= Gem::Version.new('2.0.0')
       end
 
+      def federation_2_1?
+        Gem::Version.new(federation_version.to_s) >= Gem::Version.new('2.1.0')
+      end
+
       def federation_sdl(context: nil)
         document_from_schema = FederatedDocumentFromSchemaDefinition.new(self, context: context)
 
@@ -37,8 +49,8 @@ module ApolloFederation
         output
       end
 
-      def link_namespace
-        @link ? @link[:as] : find_inherited_value(:link_namespace)
+      def default_link_namespace
+        @default_link_namespace || find_inherited_value(:default_link_namespace, DEFAULT_LINK_NAMESPACE)
       end
 
       def query(new_query_object = nil)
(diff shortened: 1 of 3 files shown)


utay:
CI fails on graphql-ruby <= 1.11, it looks like schema directives support was added in [1.12](https://github.com/rmosolgo/graphql-ruby/blob/master/CHANGELOG.md#1120-20-january-2021):
```
  1) ApolloFederation::ServiceField with older versions of GraphQL and the interpreter runtime behaves like service field returns the federation SDL with compose directives for the schema
     Failure/Error: super(*args, **kwargs, &block)
     
     ArgumentError:
       unknown keyword: :directives
```

grxy:
@utay Thanks for putting this together! I will try to take a look this week, but will be OOO for a couple weeks after in the case I'm not able to get to it this week.

spec/apollo-federation/service_field_v2_spec.rb
```diff
@@ -128,6 +128,98 @@ def execute_sdl(schema)
...
+      product = Class.new(base_object) do
+        graphql_name 'Product'
+
+        field :upc, String, null: false
+      end
+
+      query_obj = Class.new(base_object) do
+        graphql_name 'Query'
+
+        field :product, product, null: true, directives: { complexity_directive => { fixed: 1 } }
```
grxy, line 187: I think this is always going to fail on older GraphQL ruby versions. I think we have a couple options potentially:
1. Drop support for GraphQL Ruby < 1.12 (or maybe even < 2.0)
2. Only run these tests on newer versions
  - If we do this, we should add code to make sure that we warn or raise if composeDirective is used on an older GraphQL Ruby version

I lean towards the first option.

lib/apollo-federation/schema.rb
```diff
@@ -29,6 +37,10 @@ def federation_2?
         Gem::Version.new(federation_version.to_s) >= Gem::Version.new('2.0.0')
       end
 
+      def federation_2_1?
```
grxy, line 40: I don't think we should add specific minor version support. We do have some cleanup to do, but it's going to be easier to maintain this lib if we simply aim to support the latest version of federation instead of incrementally supporting minor versions of 2.x. See [this issue](https://github.com/Gusto/apollo-federation-ruby/issues/221) for more context.

spec/apollo-federation/service_field_v2_spec.rb
```diff
@@ -128,6 +128,98 @@ def execute_sdl(schema)
...
+      query_obj = Class.new(base_object) do
+        graphql_name 'Query'
+
+        field :product, product, null: true
+      end
+
+      schema = Class.new(base_schema) do
+        query query_obj
+        federation version: '2.3',
+                   links: [{
```
grxy, line 147: Can you explain the need to support multiple links in this PR? I wonder if we can do that separately and skip that for this PR, to keep it smaller.

spec/apollo-federation/service_field_v2_spec.rb
```diff
@@ -128,6 +128,98 @@ def execute_sdl(schema)
...
+
+        field :product, product, null: true, directives: { complexity_directive => { fixed: 1 } }
+      end
+
+      schema = Class.new(base_schema) do
+        query query_obj
+        federation version: '2.3',
+                   links: [{
+                     url: 'https://specs.example.com/federation/v2.3',
+                     import: [complexity_directive.graphql_name],
```
grxy, line 195: This one probably doesn't need importing with an @link since it is being defined inline.

lib/apollo-federation/schema.rb
```diff
@@ -62,14 +74,44 @@ def original_query
...
+        }
+        default_link[:as] = default_link_namespace if default_link_namespace != DEFAULT_LINK_NAMESPACE
+        [default_link, *links]
+      end
+
       def federation_2_prefix
-        federation_namespace = ", as: \"#{link_namespace}\"" if link_namespace != DEFAULT_LINK_NAMESPACE
+        schema = ['extend schema']
+
+        all_links.each do |link|
```
grxy, line 99: This printing logic is fairly complex. Can we make it easier to read?

grxy (changes requested):
I have a few concerns with how this is currently written:
1. I'm not understanding the need to support multiple links here. Can we not use `@composeDirective` with a directive defined in our schema itself?
2. I don't think we want to support code paths for multiple federation minor versions. I think we need to clean up the API for defining the version, but `'2.0'`, `2.3`, `2`, etc. essentially imply the latest 2.x version for now.

spec/apollo-federation/service_field_v2_spec.rb
utay: Per the [composeDirective docs](https://www.apollographql.com/docs/federation/federated-types/federated-directives/#composedirective):
> The directive to preserve must be defined and imported from a core specification via the @link directive. 

I think `composeDirective` would be unusable without support for multiple links

spec/apollo-federation/service_field_v2_spec.rb
utay: It looks like even if it's being defined inline it needs to be referenced via a custom `link`. For example, if I try to compose a supergraph with an inline directive that isn't imported with `link` I get:
>    UNKNOWN: Directive "@..." in subgraph "..." cannot be composed because it is not a member of a core feature 

In the [docs](https://www.apollographql.com/docs/federation/federated-types/federated-directives/#composedirective) they're doing it similarly

lib/apollo-federation/schema.rb
utay: Tried to do make it easier to read but not a ruby expert, please tell me if there's a better way :)

utay (commented):
@grxy Sorry for the delay here and thank you for your review!

spec/apollo-federation/service_field_v2_spec.rb
utay: Also if we declare a custom directive as part of the https://specs.apollo.dev link it fails. For example for:
```
 @link(
    url: "https://specs.apollo.dev/federation/v2.3"
    import: ["@shareable", "@key", "@composeDirective", "@myCustomDirective"]
  )
```
compose fails with:
```
UNKNOWN: [...] Cannot import unknown element "@myCustomDirective".
```

spec/apollo-federation/service_field_v2_spec.rb
grxy: @utay Yeah we'd probably want to do it in a separate PR and publish a new major version.

spec/apollo-federation/service_field_v2_spec.rb
grxy: Got it. Makes sense.

spec/apollo-federation/service_field_v2_spec.rb
utay: @grxy Any plan to do that in the near future? It was the main blocker here I think

sofie-c (commented):
Hi @utay - the team took a look at this yesterday and had some high-level feedback: 

- Because this change would require dropping support for other versions of GQL-Ruby, this isn't a trivial lift. We've announced in #267 that we're de-prioritizing these types of changes. 
- We'd also like to see some cleanup to this PR before it's in a merge-able state, just around the printing, hard-coding the default link url, etc.

Thank you for taking the time to open this PR and add new functionality. We'd like to eventually support the `@composeDirective` but it's something we're de-prioritizing right now.

PR #258, opened 2023-07-13
docs(changelog): add breaking change to 3.4.0

jmpage:
This adds information to v3.4.0 highlighting the breaking change to composition discussed in Gusto/apollo-federation-ruby#221 and providing guidance on how to proceed.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -148,6 +148,11 @@
 
 * add support for the [@interface](https://github.com/interface)Object directive ([#218](https://github.com/Gusto/apollo-federation-ruby/issues/218)) ([c7b987d](https://github.com/Gusto/apollo-federation-ruby/commit/c7b987de1d2b32a4a77ceb09718373ffa5a60abb))
 
+
+### NOTES
+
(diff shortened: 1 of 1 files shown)


CHANGELOG.md
```diff
@@ -148,6 +148,11 @@
 
 * add support for the [@interface](https://github.com/interface)Object directive ([#218](https://github.com/Gusto/apollo-federation-ruby/issues/218)) ([c7b987d](https://github.com/Gusto/apollo-federation-ruby/commit/c7b987de1d2b32a4a77ceb09718373ffa5a60abb))
 
+
+### BREAKING CHANGES
```
grxy, line 152: I don't think we should specifically label this as a breaking change until we decide on the path forward for federation version updates. If #222 is merged, then that would just be a fix for the breaking change bug. If we decide not to move forward with that, then I would be fine leaving this note as is, and we can aim to be more clear about this moving forward. Can you just mark this as a note for now?

jmpage: Done!

grxy (approved):

PR #263, opened 2023-09-13
fix: Adds support for defining Interface entities resolvers

moonflare:
### Context
As part of the Federation v2.3 spec, the Apollo team added a new feature, to support defining and referencing Entities that are interfaces. You can read more about this [here](https://github.com/apollographql/federation/issues/2277), and check the initial implementation [here](https://github.com/apollographql/federation/pull/2279).

This feature consists of two parts:
1. An Interface Entity defined in Subgraph A can be referenced from Subgraph B using the `@interfaceObject` directive. This work was added in https://github.com/Gusto/apollo-federation-ruby/pull/218
2. An Interface Entity can be defined in your subgraph. This part is missing and this is what this PR adds.

### Description
* Updates the Union validation so we can allow an interface to be a union member
  * The main issue here is the fact that an `Union` in `GraphQL` can't be an `Interface` according to the [spec](https://spec.graphql.org/October2021/#sec-Unions.Type-Validation), but at the same time, according to the Apollo Federation [spec](https://www.apollographql.com/docs/federation/federated-types/interfaces), an interface can be an Entity, and an Entity is an Union. Therefore, we have to extend the validation (`assert_valid_union_member`) for the Entity union to allow `Interfaces` (more precisely `Modules`) as possible types.
* Adds validation that all implementing types of an Entity Interface also have the `@key` directive. This is a requirement from the spec.
* Adds `underscore_reference_keys` support to interfaces
  * In a previous Pull Request (#248) there was added support for underscoring the keys and while this covered the `Object` class, it didn't also cover the `Interface` module. This commit adds the same `underscore_reference_keys` method to the `Interface` module.
* Adds some docs for the `Interface Entities` feature support.

cc @nogates

### DIFF
--- a/lib/apollo-federation/entity.rb
+++ b/lib/apollo-federation/entity.rb
@@ -9,5 +9,18 @@ module ApolloFederation
     def self.resolve_type(object, context)
       context[object]
     end
+
+    # The main issue here is the fact that an union in GraphQL can't be an interface according
+    # to the [spec](https://spec.graphql.org/October2021/#sec-Unions.Type-Validation), but at
+    # the same time, according to the Federation spec, an interface can be an Entity, and an Entity
+    # is an union. Therefore, we have to extend this validation to allow interfaces as possible types.
+    def self.assert_valid_union_member(type_defn)
+      if type_defn.is_a?(Module) &&
+          type_defn.included_modules.include?(ApolloFederation::Interface)
+        # It's an interface entity, defined as a module
+      else
+        super(type_defn)
+      end
+    end
   end
 end
--- a/spec/apollo-federation/entities_field_interfaces_spec.rb
+++ b/spec/apollo-federation/entities_field_interfaces_spec.rb
@@ -1,0 +1,500 @@
+# frozen_string_literal: true
+
+require 'spec_helper'
+require 'graphql'
+require 'apollo-federation/schema'
+require 'apollo-federation/field'
+require 'apollo-federation/object'
+require 'apollo-federation/interface'
+
+require_relative './spec_types'
+
+RSpec.describe ApolloFederation::EntitiesField do
+  shared_examples 'entities field' do
+    context 'when an interface with the key directive doesn\'t exist' do
+      it 'does not add the _entities field' do
+        schema = Class.new(base_schema) do
+        end
+
+        expect(schema.to_definition).to match_sdl(
+          <<~GRAPHQL,
+            type Query {
+              _service: _Service!
+            }
+
+            """
+            The sdl representing the federated service capabilities. Includes federation
+            directives, removes federation types, and includes rest of full schema after
+            schema directives have been applied
+            """
+            type _Service {
+              sdl: String
+            }
+          GRAPHQL
+        )
+      end
+    end
+
+    context 'when an interface with the key directive exists' do
+      context 'when a Query object is provided' do
+        let(:query) do
+          user_class = SpecTypes::User
+          Class.new(SpecTypes::BaseObject) do
+            graphql_name 'Query'
+            field :user, user_class, null: true
+          end
+        end
+
+        let(:schema) do
+          query_class = query
+          Class.new(base_schema) do
+            query query_class
+            orphan_types SpecTypes::AdminType, SpecTypes::EndUserType
+            def self.resolve_type(_abstract_type, _obj, _ctx)
+              raise(GraphQL::RequiredImplementationMissingError)
+            end
+          end
+        end
+
+        it 'sets the Query as the owner to the _entities field' do
+          expect(
+            schema.query
+              .fields['_entities']
+              .owner.graphql_name,
+          ).to eq('Query')
+        end
+
+        it 'adds an _entities field to the Query object' do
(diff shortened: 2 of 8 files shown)


moonflare:
Hey @grxy! Could you please have a look at this PR when time allows? 🙏🏻

grxy:
Hey there @moonflare. We are in the midst of a team shuffle so reviewing this will be delayed. Thanks for the PR and for your patience! 🙏

nogates:
> Hey there @moonflare. We are in the midst of a team shuffle so reviewing this will be delayed. Thanks for the PR and for your patience! 🙏

Hey @grxy ! 

Sorry to ask, but do you when you will have time to look at this? We are going to need this pretty soon and we would love to see the code here :)

lib/apollo-federation/entity.rb
```diff
@@ -9,5 +9,18 @@ class Entity < GraphQL::Schema::Union
...
+    # the same time, according to the Federation spec, an interface can be an Entity, and an Entity
+    # is an union. Therefore, we have to extend this validation to allow interfaces as possible types.
+    def self.assert_valid_union_member(type_defn)
+      if type_defn.is_a?(Module) &&
+         type_defn.included_modules.include?(ApolloFederation::Interface)
+        # It's an interface entity, defined as a module
+      else
+        super(type_defn)
+      end
+    end
```
sethc2, line 24: I totally get why we would do this but I was curious if you think we could tweak the implementation to avoid creating a union that is technically invalid according to the GraphQL spec.

I was thinking we keep the requirement that all entity interface implementations are entities, and continue to check that when getting the list of possible types for the entities field, but then drop it from the final list of possible types and then in the entities resolver we use the `resolve_type` method on an interface and just force that resolution to happen.

Thinking something like this, though I haven't tested to see how well this plays with GraphQL ruby's `after_lazy`.

```ruby
      maybe_lazies = grouped_references_with_indices.map do |typename, references_with_indices|
        references = references_with_indices.map(&:first)
        indices = references_with_indices.map(&:last)

        # TODO: Use warden or schema?
        type = context.warden.get_type(typename)
        if type.nil? || !(type.kind == GraphQL::TypeKinds::OBJECT || (type.kind == GraphQL::TypeKinds::Interface && has_key_directive(type)))
          raise "The _entities resolver tried to load an entity for type \"#{typename}\"," \
                ' but no entity type of that name was found in the schema'
        end
        
        # ... EXISTING CODE calling resolve reference(s) ...
        
           final_result[i] = context.schema.after_lazy(result) do |resolved_value|
              # force the type to the concrete type
              type_to_set = if type.kind == GraphQL::TypeKinds::Interface
                type_class.resolve_type(resolved_value, context)
              else
                type
              context[resolved_value] = type_to_set
              resolved_value
            end
```

Do you think something like that would work?

sethc2 (commented):
Hey sorry we've been so long getting to this. Our team had a big shuffle, and Grex switched to a different team.

I had one comment that I'm curious to know if you think could work as a way to avoid adding the interface to the list of entity types.

lib/apollo-federation/entity.rb
```diff
@@ -9,5 +9,18 @@ class Entity < GraphQL::Schema::Union
     def self.resolve_type(object, context)
       context[object]
     end
+
+    # The main issue here is the fact that an union in GraphQL can't be an interface according
+    # to the [spec](https://spec.graphql.org/October2021/#sec-Unions.Type-Validation), but at
+    # the same time, according to the Federation spec, an interface can be an Entity, and an Entity
+    # is an union. Therefore, we have to extend this validation to allow interfaces as possible types.
```
sofie-c, line 16: nit: 

```suggestion
    # The main issue here is the fact that a union in GraphQL can't be an interface according
    # to the [spec](https://spec.graphql.org/October2021/#sec-Unions.Type-Validation), but at
    # the same time, according to the Federation spec, an interface can be an Entity, and an Entity
    # is a union. Therefore, we have to extend this validation to allow interfaces as possible types.
```

spec/apollo-federation/entities_field_interfaces_spec.rb
sofie-c, line 1: I'm not a fan that this behavior is tested in a separate file. In my initial review of this PR, I missed these tests because I was looking for them in `entities_field_spec`. Since all the logic lives in `entities_field.rb` I think I'd like to see all of the test cases describing its behavior in one file. Can you elaborate on your decision to make this a separate file?

sofie-c (commented):
Thanks for your patience as we restructured the team and took a while to get to this PR. I gave this an initial look but am still trying to understand all the context around the change. 

Left some initial nits / question. I think we as a team want to discuss some more around Seth's question and make sure we are aligned with the proposed solution in this PR.

nogates:
Thanks @sofie-c ! I will have a look this week or early next week :)

craig-day:
Hey @sofie-c and @sethc2 after some moving around on our side as well, I'm going to pick this work up from @moonflare. Do we still want to keep things in this PR, or would you be OK with me opening up a new PR based on the latest main?

PR #265, opened 2023-10-06
feat: Add entity interface support

rosscooperman:
This is a super minimal implementation of support for interfaces as entities. Really just enough to get our `@interfaceObject` + `@key` on interface setup working.

I don't expect this is anywhere near ready to merge but was hoping to get some feedback on the approach before I go further. Some things I know I will need to do (but didn't need for our specific purposes):
1. Support `underscore_reference_keys` on interfaces. We don't happen to use it so I just guarded against calling it if it's not there but will need to add support to the `ApolloFederation::Interface` module.
2. Will need to support `resolve_type` on the interface as opposed to the schema. Need to look into whether there's a consistent way in `graphql-ruby` to resolve an interface type that will have the framework check all relevant options -- probably is, just didn't look closely because we resolve all of our interface types through `Schema.resolve_type`.
3. Tests.

### DIFF
--- a/lib/apollo-federation/entities_field.rb
+++ b/lib/apollo-federation/entities_field.rb
@@ -71,7 +71,11 @@ module ApolloFederation
               # calls return the same object, it might not have the right type
               # Right now, apollo-federation just adds a __typename property to the result,
               # but I don't really like the idea of modifying the resolved object
-              context[resolved_value] = type
+              if type.kind == GraphQL::TypeKinds::INTERFACE
+                context[resolved_value] = context.schema.resolve_type(type, resolved_value, context)
+              else
+                context[resolved_value] = type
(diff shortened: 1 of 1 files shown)


slauppy (commented):
Thanks for working on this! Did you see [this other PR ](https://github.com/Gusto/apollo-federation-ruby/pull/263) where we're also talking about entity interface support? How do you feel about collaborating on that PR instead, since it's further along?

rosscooperman:
> Thanks for working on this! Did you see [this other PR ](https://github.com/Gusto/apollo-federation-ruby/pull/263) where we're also talking about entity interface support? How do you feel about collaborating on that PR instead, since it's further along?

Sounds good to me!

PR #270, opened 2023-12-14
fix: ensure query object is traversed

erikkessler1:
## Problem

The way we override the `Schema#query` method makes it so that we don't call `add_type_and_traverse` when setting the new query type like graphql-ruby does [here](https://github.com/rmosolgo/graphql-ruby/blob/5eac86139138332d17d4eb5ea3f59ae8e0aa8cfb/lib/graphql/schema.rb#L416). Currently we only add and traverse the query type when accessing the query.

The following script demonstrates this:

```ruby
require 'bundler/inline'

gemfile do
  source 'https://rubygems.org'
  gem 'apollo-federation'
end

class CatType < GraphQL::Schema::Object
end

class QueryType < GraphQL::Schema::Object
  field :cat, CatType, null: false
end

class Schema < GraphQL::Schema

  query QueryType
end

p Schema.get_type('Cat')
#=> nil
```

## Solution

This PR makes it so that we call `super(federation_query_object)` when the new query object is being set.

### DIFF
--- a/lib/apollo-federation/schema.rb
+++ b/lib/apollo-federation/schema.rb
@@ -42,18 +42,15 @@ module ApolloFederation
       end
 
       def query(new_query_object = nil)
-        if new_query_object
-          @orig_query_object = new_query_object
-        else
-          if !@federation_query_object
-            @federation_query_object = federation_query(original_query)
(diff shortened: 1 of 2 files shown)


sethc2 (approved):
Thanks for fixing this.

PR #317, opened 2026-09-23
fix: Query._entities crashing under GraphQL::Execution::Next

fuzzyrichie:
## Summary
`field(:_entities, ...)` in `EntitiesField#define_entities_field` has no execution-mode annotation. Under `GraphQL::Execution::Next`'s default `:direct_send` mode, that means the field is resolved by calling the raw backing object directly instead of the wrapped Query type instance — and for a root Query field, that backing object is `root_value`, `nil` by default. This produces:

```
NoMethodError: undefined method '_entities' for nil
```

Classic execution never hits this, since it dispatches against the wrapped Query object (which has `_entities` mixed in via `EntitiesField`), regardless of what the backing object is.

Found via a production incident at Gusto migrating a subgraph to `GraphQL::Execution::Next` — this fires on any cross-subgraph entity reference resolution once Next is enabled, independent of anything else in the schema.

## Fix
- Adds `resolve_static: true` to the `_entities` field declaration, **feature-detected** against `GraphQL::Schema::Field#initialize`'s parameters rather than assumed, since this gem supports graphql-ruby versions that predate `GraphQL::Execution::Next` entirely.
- Moves the entity-resolution logic into a class method (`ClassMethods#_entities(context, representations:)`), matching `resolve_static`'s dispatch signature (`owner_class.public_send(key, context, **args)`). The instance method now delegates to it, so classic execution (which ignores `resolve_static` and falls back to instance-method dispatch) is unaffected.
- Changes the final `GraphQL::Execution::Lazy.all(...).then { ... }` to `.value`. Next's list-field result handling calls `.size` directly on whatever the resolver returns for a list field, without lazy-unwrapping it first — so returning a raw classic `Lazy` object (correct under classic) crashes there too. `.value` fully resolves it before returning.

## Test plan
- [x] Existing `entities_field_spec.rb` suite (27 examples) — unchanged, passes under both the pinned graphql 2.0.31 and graphql 2.6.11 (identical pass/fail set with and without this diff on 2.6.11 — the pre-existing 2.6.x failures elsewhere in the repo, e.g. `service_field_v2_spec.rb`'s `orphan_types`, are unrelated to this change)
- [x] Full repo suite (151 examples) — 0 failures, rubocop clean, both against the pinned graphql 2.0.31
- [x] New guarded spec (`under GraphQL::Execution::Next`, skipped when `GraphQL::Execution::Next` isn't defined) proves classic and Next now return identical results for `_entities`, including a real `resolve_reference` implementation — verified manually against graphql 2.6.11 (this repo's CI still pins 2.0.31, so the new spec won't run there until that's bumped, but it's guarded to activate automatically when it is)

🤖 Generated with [Claude Code](https://claude.com/claude-code)


sethc2 (commented):
Looks good to me. (Letting graphql team do actually approval)

lib/apollo-federation/entities_field.rb
```diff
@@ -20,80 +27,90 @@ def define_entities_field(possible_entities)
...
 
-      # Make sure we've resolved the outer level of lazies so we can return an array with a possibly lazy
-      # entry for each requested entity
-      GraphQL::Execution::Lazy.all(maybe_lazies).then do
-        final_result
+        # Make sure we've resolved the outer level of lazies so we can return an array with a possibly lazy
+        # entry for each requested entity
+        GraphQL::Execution::Lazy.all(maybe_lazies).then do
+          final_result
+        end.value
```
slauppy: Nit: this is only needed when the schema doesn't register its own `lazy_resolve`. Next's `resolves_lazies?` is false then, so the `Lazy` is never unwrapped. It also means classic now syncs promises inside the resolver. That's harmless for `_entities`, but `return final_result if maybe_lazies.none? { |l| l.is_a?(GraphQL::Execution::Lazy) }` before the `Lazy.all` would keep classic's behavior exactly. Either way, a short comment on why it's here would help. FYI, per-item `Lazy`/`Dataloader::Request` values still fail under Next on schemas with no custom `lazy_resolve`. That's a graphql-ruby limitation, not something this PR introduced.

spec/apollo-federation/entities_field_spec.rb
```diff
@@ -617,4 +617,90 @@ def self.resolve_reference(reference, _context)
       end
     end
   end
+
+  # GraphQL::Execution::Next didn't exist before graphql-ruby 2.6; guarded rather than pinned to
+  # this repo's own (older) Gemfile.lock version so it activates automatically once that's bumped.
+  if defined?(GraphQL::Execution::Next)
```
slauppy: CI pins `graphql ~> 2.0.0`, so this block never runs in CI. Could you add a matrix entry on graphql 2.6.x, even just for this file? Otherwise a later change could break Next with nothing going red.

spec/apollo-federation/entities_field_spec.rb
```diff
@@ -617,4 +617,90 @@ def self.resolve_reference(reference, _context)
...
+              ... on TypeWithKey {
+                id
+                otherField
+              }
+            }
+          }
+        GRAPHQL
+      end
+
+      let(:representations) { [{ __typename: 'TypeWithKey', id: 123 }] }
```
slauppy: Nit: could the Next block also cover interleaved `__typename`s, an unresolvable representation, a `resolve_references` type, and a schema with `use GraphQL::Batch`? They pass locally on this branch, so these would just lock that in.

swcollard (commented):
_I am a Claude agent working on behalf of Sam (@swcollard) from the GraphQL team._

Two items:

1. **`_service` has the same problem.** `lib/apollo-federation/service_field.rb:16` defines `_service` as a plain `field(:_service, Service, null: false)`, and the value comes from an instance method. Under `GraphQL::Execution::Next`, it will fail the same way `_entities` did (`undefined method '_service' for nil`). Please give it the same `resolve_static` treatment, in this PR or in a follow-up.

2. **CI does not run the new Next specs.** They are inside `if defined?(GraphQL::Execution::Next)` (`spec/apollo-federation/entities_field_spec.rb:623`), and the Gemfile pins `graphql ~> 2.0.0`. So in CI the describe block is not registered, and nothing guards this fix from regressions. Can you add a CI job (or an appraisal Gemfile) that runs the specs on graphql 2.6.x?

fuzzyrichie:
@swcollard (via Sam's review) — both addressed:

1. **`_service` had the identical problem.** Gave it the same `resolve_static: true` treatment as `_entities` (feature-detected via a new shared `ApolloFederation::RESOLVE_STATIC_SUPPORTED` constant, used by both now). Tracing this through end-to-end also surfaced a second bug one level down: `Service#sdl`'s plain `field(:sdl, ...)` had no `hash_key:`, so it hit the same `:direct_send` gap on the `{ sdl: ... }` Hash `_service` resolves to — fixed that too, since the `_service` fix would've been incomplete otherwise.

2. **CI now runs the Next specs.** Added `gemfiles/graphql_2.6.gemfile` (pinning `graphql ~> 2.6.7`) and a new `ruby-test-graphql-next` CI job that runs `rspec --tag next_execution` against it across the same Ruby matrix (3.2/3.3/3.4), wired into the `Release` job's `needs:`. Deliberately scoped to just the tagged specs rather than the full suite — several unrelated specs (schema SDL printing, `orphan_types`) aren't yet compatible with graphql 2.6.x, which is a separate, pre-existing gap from the one this PR closes.

(One follow-up bug of my own along the way: the first version of that lockfile was resolved under Ruby 3.4 and pulled in `rbs 4.2.0`, which requires Ruby >= 3.3 — broke the 3.2 leg of the new job. Removed the committed lockfile so each Ruby version resolves its own compatible deps independently; verified locally on 3.2 and 3.4 both.)

All CI checks are green now, including the new graphql-2.6 job across all three Ruby versions.

